PR #5, opened 2023-07-28
Use ACCESS-OM2 grid, topography and initial conditions

dougiesquire:
This PR moves the configuration to ACCESS-OM2 1 deg grid, topography, and initial conditions - see https://github.com/COSIMA/access-om3/issues/36. A Python script is also included for generating ESMF mesh files - this should possibly be moved elsewhere at some point?

A few things to note. If everyone's okay with it, I'll leave these to be addressed in subsequent PRs. I'll open issues for them once this PR is merged.

- The MOM temperature and salt initial conditions used by ACCESS-OM2 are conservative temperature and practical salinity. Here, we use these initial conditions as is, but use the MOM6 `"WRIGHT"` EOS, so treat them as though the temperatures are actually potential temperatures - see https://github.com/COSIMA/access-om3/issues/20#issuecomment-1644991506.

- ~The temperature and salt initial conditions are on the same grid as the model grid, which means we _should_ be able to run with `TEMP_SALT_INIT_VERTICAL_REMAP_ONLY = True` in `MOM_input`. However, I couldn't get this to work. We should revisit why this is~. This works using an access-om3 executable, rather than the cime-built one I was playing with early in this PR. Possibly due to FMS version. 

- I'm currently generating an ESMF mesh file for the MOM grid and providing that. This is problematic as FMS2 doesn't support meshes (we're currently using FMS1). The MOM6 NUOPC cap does provide a option to set `use_mommesh = .false.` which then generates an ESMF Grid internally from the MOM grid. However, this wasn't trivial to get working in this configuration. It is being used in other configurations, so it shouldn't be hard to sort out.

- I've unset `rof2ocn_ice_rmapname` and `rof2ocn_liq_rmapname` in `nuopc.runconfig`. This should mean these mappings are generated at run-time, but for efficiency we should probably pre-compute these (and other mappings).

- The ice initial conditions are generated from a 3hr run of the [1deg_jra55_ryf configuration](https://github.com/COSIMA/1deg_jra55_ryf/tree/7c6be951530901082be8106a4f819e437fbeb19b). This probably doesn't require a follow-up issue, but noting anyway. See https://github.com/COSIMA/access-om3/issues/50.

- I've removed the step of remapping atm/rof forcing to an intermediate grid by setting the data model grids equal to the ocn/ice grid. This means the spatial remapping is done in CDEPS prior to temporal interpolation, which should be more efficient that performing the spatial remapping with CMEPS after time interpolation. However, it's not obviously possible atm to pass precomputed remapping weights to CDEPS like you can for CMEPS remapping. We should consider adding this as it will presumably make a big difference for the high resolution configurations. We should also confirm that CMEPS recognises when the source and destination grids are the same and doesn't try to remap in this case.

- I haven't yet made any effort to get the parameter settings the same between the ACCESS-OM2 and OM3 configurations.


dougiesquire:
Apologies, the commit history is a bit of a mesh

aekiss:
Awesome, thanks @dougiesquire I'll look at this next week

ice_in
```diff
@@ -21,31 +21,33 @@
   pointer_file = "./rpointer.ice"
   print_global = .true.
   print_points = .false.
+  restart = .true.
```
ezhilsabareesh8, line 24: For the first run, is this flag needs to set to be`.false.`? Refer this [issue](https://github.com/payu-org/payu/issues/349#issue-1825453475)

ice_in
```diff
@@ -21,31 +21,33 @@
...
   restart_ext = .false.
   restart_file = ""
   restart_format = "default"
+  runtype = "initial"
```
ezhilsabareesh8, line 28: As above, [issue 349](https://github.com/payu-org/payu/issues/349#issue-1825453475)

ezhilsabareesh8 (commented):
Thanks @dougiesquire. Looks good to me.

config.yaml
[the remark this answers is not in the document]
dougiesquire: I'm not sure... I haven't changed this from `main`.

ice_in
dougiesquire: I don't think so? But with the CESM NUOPC driver it's not very clear which of the multiple seemingly-equivalent restart flags are are used and which are ignored. Thanks for pointing this out - I'll try to get to the bottom of this today.

Thanks also for pointing out the payu issue - I'd missed that

dougiesquire: I looked into this and in the CICE NUOPC cap the `start_type` set in `nuopc.runconfig` is used over the `runtype` set in `ice_in`. Basically (see [here](https://github.com/ESCOMP/CICE/blob/34dc66707f6b691b1689bf36689591af3e8df270/cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90#L404)):

- If `start_type` is `"continue"` or `"branch"`, then CICE parameters `restart = .true.`, `runtype = "continue"` and `use_restart_time = .true.` are set;
- If `start_type` is `"startup"` then CICE parameter `runtype = "initial"` is set.

So I think everything should work correctly as I've implemented it. But it's probably clearer to remove `runtype = "initial"` from `ice_in`.

config.yaml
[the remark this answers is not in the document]
dougiesquire: Did you want me to change something here @ezhilsabareesh8?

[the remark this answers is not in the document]
ezhilsabareesh8: I'm not sure that the `parallelio/2.5.10-intel-2021.6.0` is required or not in `config.yaml`. In `main` branch it is there [(Reference)](https://github.com/COSIMA/MOM6-CICE6/blob/fca7fafcacb11cb4faae9660ae81be834a7dfc79/config.yaml#L43)

ice_in
ezhilsabareesh8: Thanks @dougiesquire

config.yaml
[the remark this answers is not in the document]
dougiesquire: A question for @micaeljtoliveira I guess? Is there a reason why you flagged this in the first place?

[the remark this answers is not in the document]
micaeljtoliveira: Yes, in theory it is needed.

micaeljtoliveira:
If it's okay with everyone, I would to wait a bit before merging this (and also #6). The reason is that I would like to tag a version of then main branch that corresponds to the "unchanged" CESM configuration as it was generated by @dougiesquire . The idea would be to also tag a version of the model that exactly reproduces the results obtained with the corresponding CESM executable. This would provide a clear starting point for all future developments and changes.

dougiesquire:
Note, I've added some info to the wiki on how the driver initialisation parameters interact with the component initialisation parameters: https://github.com/COSIMA/access-om3/wiki/NUOPC-driver#component-initialisation

aekiss:
Great, thanks @dougiesquire

dougiesquire:
@aekiss, this is now failing on "continue" runs, so probably best to hold off reviewing until I've worked out why (I'm not exactly sure when this problem started, as most of my testing has just been to run a single "startup" run).

dougiesquire:
> But it's probably clearer to remove runtype = "initial" from ice_in

Just noting that I think this is actually needed. It is important to let the NUOPC cap set this parameter based on the NUOPC `start_type` - see https://github.com/COSIMA/access-om3/wiki/NUOPC-driver#cice6.

dougiesquire:
MOM is crashing on day 22 of the second monthly run. This crash still occurs if I change the atm and rof model grids back to the original (192x94 and 360x180) grids, but it does not occur if I use ice initial conditions `ice_ic = "default"` rather than the restart I created from ACCESS-OM2. That said, I don't think there's anything wrong with these initial conditions as it makes it 21 days before crashing.

The MOM error is:
```
WARNING from PE    47: Extreme surface sfc_state detected: i= 341 j= 280 lon=  67.145 lat=  70.619 x=  60.500 y=  80.743 D= 1.1806E+01 SSH= 8.8014E+00 SST=-2.1004E+00 SSS= 3.8733E+01 U-= 2.3777E-02 U+= 0.0000E+00 V-= 3.8195E-02 V+= 0.0000E+00

FATAL from PE     0: There were a total of         1 locations detected with extreme surface values!
```

If I'm interpreting this correctly, this suggests a SSH of nearly 9m. The location of the issue is in the Kara Sea (TIL about the Kara Sea). See the magenta cross:

Screenshot 2023-08-04 at 3 11 25 pm

@aekiss the MOM baroclinic timestep is already quite a bit shorter than it is in ACCESS-OM2 (1800 s vs 5400 s). I presume we don't want to reduce this further (ideally would we increase it?). Any thoughts on the best way forward?

dougiesquire:
The issue still occurs (though 1 day later) after adding salinity restoring as in ACCESS-OM2.

aekiss:
Hm, interesting. Is there anything unusual in the sea ice or wind at that location and time?

aekiss:
Agreed that we'd ultimately like a longer baroclinic timestep, but does a shorter timestep fix the issue?

aekiss:
We used Rayleigh damping in a few select locations in OM2 (see sec 3.2.7 [here](https://cosima.org.au/wp-content/uploads/2023/07/ACCESS-OM2-1-025-010deg-2023-06-30-b26d655.pdf)). At 0.1° (but not 1°) we needed damping in the nearby Kara Strait. But that's really an engineering fix, best avoided if we can.

aekiss:
Is there anything odd in the topography in that region?

MOM_input
```diff
@@ -106,6 +26,28 @@ HFREEZE = 10.0
                                 ! where OBLD is the boundary layer depth. If HFREEZE <= 0 (default),
                                 ! melt potential will not be computed."
 
+C_P = 3992.0
```
aekiss, line 29: We'll need to be careful with setting this once we have finalised the equation of state. OM2 uses `CP_OCEAN_PRETEOS10 = 3992.10322329649d0` but that's specific to the pre-TEOS10 EOS. The `ROQUET` value would presumably be the MOM6 default, which matches `CP_OCEAN_TEOS10    = 3991.86795711963d0` in MOM5. We also need to make sure CICE6 uses the same value. See https://github.com/COSIMA/access-om3/issues/53

aekiss: Actually, it looks like CICE6 doesn't use ocean heat capacity.

aekiss: is `ENABLE_THERMODYNAMICS` true?

dougiesquire: Yes, `ENABLE_THERMODYNAMICS` is true by default

MOM_input
```diff
@@ -154,48 +89,63 @@ SAVE_INITIAL_CONDS = True
...
+                                ! "The total number of thickness grid points in the
+                                ! y-direction in the physical domain. With STATIC_MEMORY_
+                                ! this is set in MOM_memory.h at compile time.
+                                ! The number of processors in the x-direction. With
+                                ! STATIC_MEMORY_ this is set in MOM_memory.h at compile time.
+                                ! The number of processors in the x-direction. With
+                                ! STATIC_MEMORY_ this is set in MOM_memory.h at compile time.
+                                ! The processor layout that was acutally used."
+
+IO_LAYOUT = 1, 1
```
aekiss, line 138: we'd want to adjust this for a production run, e.g. https://github.com/COSIMA/1deg_jra55_ryf/blob/e9796f2c1617cbc47ba0e1ab25c30fc084779aff/ocean/input.nml#L230-L231

aekiss:
I don't think we need to worry about fixing the crash in the Kara Sea before we merge this, as crashes like this will need ongoing tweaks to resolve, and we won't be running with this forcing anyway.

I did a test merge of this branch with [Support-JRA55-do](https://github.com/COSIMA/MOM6-CICE6/tree/Support-JRA55-do)
here: https://github.com/COSIMA/MOM6-CICE6/tree/om3-test

This ran much longer (until 13 Oct in the first year), then failed at the head of the Persian Gulf with
```
WARNING from PE    31: Extreme surface sfc_state detected: i= 329 j= 194 lon=  48.500 lat=  26.524 x=  48.500 y=  26.524 D= 1.1806E+01 SSH= 1.0595E+01 SST= 2.5510E+01 SSS= 4.5002E+01 U-= 0.0000E+00 U+=-4.1033E-02 V-= 0.0000E+00 V+= 1.8850E-02
```
There's nothing unusual in the winds at that location/time. I'll investigate whether changing the barotropic split (`DTBT`) helps.

MOM_input
dougiesquire: I've added a reminder about setting Cp to https://github.com/COSIMA/access-om3/issues/20

MOM_input
dougiesquire: Could we think about this in a separate PR? I've opened an issue here: https://github.com/COSIMA/access-om3/issues/54

dougiesquire:
Apologies, I was working on other things for most of last week. I'm happy to merge if others are.

aekiss:
I think @micaeljtoliveira still wants to [tag a CESM branch](https://github.com/COSIMA/MOM6-CICE6/pull/5#issuecomment-1661276964) first

aekiss:
I've made a separate issue to record attempts to resolve the crashing issue https://github.com/COSIMA/access-om3/issues/55

dougiesquire:
There have been recent updates to the CICE NUOPC cap that add an additional check that the mask in the supplied mesh matches an internally generated mesh. After rebasing onto `1deg_jra55do_ryf`, the configuration in this PR currently fails that check. I've been trying to explore why this is and I'll include details below for the purpose of having things documented in the open.

dougiesquire:
From what I can tell, the masks are the same between the supplied mesh and `kmt.nc` files. [This](https://gist.github.com/dougiesquire/232b71b6d7629b2d0ae7fffbdc724d6a) notebook compares mesh and kmt masks for both the displaced pole grid in [`1deg_jra55do_ryf`](https://github.com/COSIMA/MOM6-CICE6/tree/e2de7a8489db2ad182ab77b1566cb3ddc53569df) (which runs successfully) and the tripolar grid in this PR (which doesn't run). The mesh and kmt masks match in both cases.

(Note, this notebook also includes some potentially useful functions for reading the CICE binary input files and for converting from ULON/ULAT to TLON/TLAT)

dougiesquire:
The implementation of the check doesn't make sense to me (see [here](https://github.com/ESCOMP/CICE/blob/8e2aab217ece5fae933a1f2ad6e0d6ab81ecad8a/cicecore/drivers/nuopc/cmeps/ice_mesh_mod.F90#L667-L689) for the code). The start and stop values of the `do` loops over `i` and `j` indexes are changed within each loop. As far as I know, in Fortran the number of iterations in a loop is fixed on commencement of the loop, so updating the start/stop values does nothing.

I've compiled a version of access-om3 that implements the mesh check in way that makes more sense to me - see [here](https://github.com/COSIMA/access-om3/compare/main...dougiesquire:access-om3:reimplement_mesh_check#diff-8123e896b9d2e25c316b61bbd61e0f6d22dc4b70bbf4c7f76ab8641edce41497R667-R689). With this change, both the displaced pole configuration and tripolar configuration (this PR) pass the check.

@micaeljtoliveira does the code linked in the first paragraph make any sense to you, or do you agree this looks like a bug? (I still don't understand why the displaced pole configuration passes the mesh checks using the original code...)

dougiesquire:
What's weird is that the code has [been in CICE for 10 months](https://github.com/ESCOMP/CICE/commit/9808b5195b2af77d2b70f5fec3e417b0a6020c5f#diff-368a66316c59c3283156c7051e3037f64aec1ac6fa3eec9a0e510f306fddd34cR667-R689) and no one else seems to have had any issues...

micaeljtoliveira:
@dougiesquire I agree, the code looks wrong and your changes should fix it.

I'll patch the code so we can move forward, but this should be reported upstream.

micaeljtoliveira:
> What's weird is that the code has [been in CICE for 10 months](https://github.com/ESCOMP/CICE/commit/9808b5195b2af77d2b70f5fec3e417b0a6020c5f#diff-368a66316c59c3283156c7051e3037f64aec1ac6fa3eec9a0e510f306fddd34cR667-R689) and no one else seems to have had any issues...

Yes, very weird!

dougiesquire:
Thanks for taking a look. I'll open an issue in the CICE repo

dougiesquire:
> I'll patch the code so we can move forward

Do you mind if I add the patch? I'd like to go through the new process if possible

micaeljtoliveira:
Oh, sorry, already done: https://github.com/COSIMA/access-om3/pull/63

micaeljtoliveira:
@dougiesquire I've now finished rebasing this branch. Can I force push to your branch once #6 is merged? In any case, you can see how it looks here: https://github.com/COSIMA/MOM6-CICE6/tree/om2_grid and these are the changes: https://github.com/COSIMA/MOM6-CICE6/compare/Support-JRA55-do...om2_grid

dougiesquire:
> Can I force push to your branch once #6 is merged?

Go ahead

micaeljtoliveira:
> > Can I force push to your branch once https://github.com/COSIMA/MOM6-CICE6/pull/6 is merged?
>
> Go ahead

@dougiesquire Done. Would you mind having a look and check I didn't mess up something?

dougiesquire (commented):
Ran this for 1 day without issue. A couple of questions:

- I know we discussed this in the recent TWG meeting, but there's a huge amount of input data that isn't actually being used by this configuration. This muddies the manifests and makes it difficult to work out which inputs are required and which aren't. Do you mind if I push a change to `config.yaml` to make the inputs a little more explicit?

- I originally reorganised `MOM_input` for convenience in finding/understanding parameters, but is the added convenience outweighed by added frustration when trying to incorporate upstream changes/additions?

micaeljtoliveira:
> I see you've copied the input files I'm using from ACCESS-OM2 into /g/data/ik11/inputs/access-om3/0.x.0, rather than symlinking their original location in /g/data/ik11/inputs/access-om2. Can I change these to symlinks to keep track of their provenance?

I discussed this with @aekiss last week and we agreed it would be better to fully separate the two sets of inputs (OM2 will be around for a long time, but probably not as long as OM3). Instead, we should add a README explaining the provenance of the files. We should probably do the same for the cime inputs by the way. But happy to discuss this further.

> I originally reorganised MOM_input for convenience in finding/understanding parameters, but is the added convenience outweighed by added frustration when trying to incorporate upstream changes/additions?

I think it should be okay. The number of changes/additions from upstream that can/should be added to our configurations for any given CESM update is probably going to be small. so it can be done manually (ie. with a copy-paste instead of applying a patch). Probably it's actually better to do it manually, as someone should review them carefully.

dougiesquire:
> We should probably do the same for the cime inputs by the way. But happy to discuss this further.

Just noting that this PR actually does not use the cime inputs at all, but they are probably still needed for other configs so I agree we (I) should make note of their provenance

micaeljtoliveira:
> Just noting that this PR actually does not use the cime inputs at all

Great! I was actually wondering about that, but forgot to ask.

aekiss:
The `MOM_input` changes are hard to follow as the order has also been rearranged. For the record, here are all the semantic differences as reported by [nmltab](https://github.com/aekiss/nmltab):
```
* &null  adjust_net_srestore_to_zero                                                MOM_input_1deg_jra55do_ryf
* &null  adjust_net_srestore_to_zero         True                                   MOM_input_om2_grid_iss36
* &null  ale_coordinate_config               'FILE:ocean_vgrid.nc,dz'               MOM_input_1deg_jra55do_ryf
* &null  ale_coordinate_config               'FILE:ocean_vgrid.nc,interfaces=zeta'  MOM_input_om2_grid_iss36
* &null  diag_coord_def_z                    'WOA09'                                MOM_input_1deg_jra55do_ryf
* &null  diag_coord_def_z                    'FILE:ocean_vgrid.nc,interfaces=zeta'  MOM_input_om2_grid_iss36
* &null  eqn_of_state                                                               MOM_input_1deg_jra55do_ryf
* &null  eqn_of_state                        'WRIGHT'                               MOM_input_om2_grid_iss36
* &null  fluxconst                           0.5                                    MOM_input_1deg_jra55do_ryf
* &null  fluxconst                           0.11                                   MOM_input_om2_grid_iss36
* &null  grid_file                           'ocean_hgrid_230424.nc'                MOM_input_1deg_jra55do_ryf
* &null  grid_file                           'ocean_hgrid.nc'                       MOM_input_om2_grid_iss36
* &null  mask_srestore_under_ice                                                    MOM_input_1deg_jra55do_ryf
* &null  mask_srestore_under_ice             False                                  MOM_input_om2_grid_iss36
* &null  max_delta_srestore                                                         MOM_input_1deg_jra55do_ryf
* &null  max_delta_srestore                  0.5                                    MOM_input_om2_grid_iss36
* &null  niglobal                            320                                    MOM_input_1deg_jra55do_ryf
* &null  niglobal                            360                                    MOM_input_om2_grid_iss36
* &null  njglobal                            384                                    MOM_input_1deg_jra55do_ryf
* &null  njglobal                            300                                    MOM_input_om2_grid_iss36
* &null  nk                                  60                                     MOM_input_1deg_jra55do_ryf
* &null  nk                                  50                                     MOM_input_om2_grid_iss36
* &null  restore_salinity                    False                                  MOM_input_1deg_jra55do_ryf
* &null  restore_salinity                    True                                   MOM_input_om2_grid_iss36
* &null  salt_restore_file                                                          MOM_input_1deg_jra55do_ryf
* &null  salt_restore_file                   'salt_sfc_restore.nc'                  MOM_input_om2_grid_iss36
* &null  srestore_as_sflux                                                          MOM_input_1deg_jra55do_ryf
* &null  srestore_as_sflux                   True                                   MOM_input_om2_grid_iss36
* &null  temp_salt_init_vertical_remap_only                                         MOM_input_1deg_jra55do_ryf
* &null  temp_salt_init_vertical_remap_only  True                                   MOM_input_om2_grid_iss36
* &null  temp_salt_z_init_file               'WOA05_pottemp_salt.nc'                MOM_input_1deg_jra55do_ryf
* &null  temp_salt_z_init_file               'ocean_temp_salt.res.nc'               MOM_input_om2_grid_iss36
* &null  topo_file                           'ocean_topog_230424.nc'                MOM_input_1deg_jra55do_ryf
* &null  topo_file                           'topog.nc'                             MOM_input_om2_grid_iss36
* &null  tripolar_n                          False                                  MOM_input_1deg_jra55do_ryf
* &null  tripolar_n                          True                                   MOM_input_om2_grid_iss36
* &null  use_contemp_abssal                                                         MOM_input_1deg_jra55do_ryf
* &null  use_contemp_abssal                  False                                  MOM_input_om2_grid_iss36
* &null  z_init_file_ptemp_var               'PTEMP'                                MOM_input_1deg_jra55do_ryf
* &null  z_init_file_ptemp_var               'temp'                                 MOM_input_om2_grid_iss36
* &null  z_init_file_salt_var                'SALT'                                 MOM_input_1deg_jra55do_ryf
* &null  z_init_file_salt_var                'salt'                                 MOM_input_om2_grid_iss36
```

aekiss (approved):
looks good to me, thanks @dougiesquire

PR #6, opened 2023-07-28
Support jra55 do

ezhilsabareesh8:
This PR uses the JRA55-do forcing similar to ACCESS-OM2 - Refer [issue 33](https://github.com/COSIMA/access-om3/issues/33)


dougiesquire (changes requested):
Thanks @ezhilsabareesh8. I’m running a test run at the moment. In the meantime, a few questions/comments.

- I see that our run-off component is using the “copyall” data mode. Do you know what this does, and is it what we want?
- Do we want to / can we remove the "presaero.clim_2000" and "presndep.clim_2000" data streams?

config.yaml
```diff
@@ -11,25 +12,28 @@ shortpath: /scratch/v45
...
 model: access-om3
 
-exe: /g/data/ik11/inputs/access-om3/bin/access-om3-MOM6-CICE6-2b52198
+
+exe: /g/data/ik11/inputs/access-om3/bin/access-om3-MOM6-CICE6-2b52198 
+
 input: 
     - /g/data/ik11/inputs/cime/inputdata/2023-03-10/share/meshes/T62_040121_ESMFmesh.nc # datm meshfile
-    - /g/data/ik11/inputs/cime/inputdata/2023-03-10/share/meshes/TL319_151007_ESMFmesh.nc # datm JRA stream meshfile
+    - /g/data/tm70/ds0092/model/inputs/access-om3/JRA55do-ESMFmesh.nc  # datm JRA stream meshfile
```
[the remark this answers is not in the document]
ezhilsabareesh8: Updated the file path

ice_in
```diff
@@ -198,7 +198,7 @@
   maskhalo_bound = .true.
   maskhalo_dyn = .true.
   maskhalo_remap = .true.
-  max_blocks = 8
+  max_blocks = 800
```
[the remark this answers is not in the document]
ezhilsabareesh8: Updated max_blocks to 8

ezhilsabareesh8 (commented):
Thanks @dougiesquire.  I have made all the changes.

Regarding the `copyall` data mode in `drof_in`, the only data  available for run off components is `copyall`([Refer `rof_comp_nuopc.f90`)](https://github.com/ESCOMP/CDEPS/blob/f027aa64285fb9ddad9be5c5837a6e6e279e6051/drof/rof_comp_nuopc.F90#L236-L241). The streams are defined in the xml file. 

Regarding "presaero.clim_2000" and "presndep.clim_2000" data streams, I think we need aerosol and lnd input files.

dougiesquire:
> Regarding the `copyall` data mode in `drof_in`, the only data available for run off components is `copyall`([Refer `rof_comp_nuopc.f90`)](https://github.com/ESCOMP/CDEPS/blob/f027aa64285fb9ddad9be5c5837a6e6e279e6051/drof/rof_comp_nuopc.F90#L236-L241). The streams are defined in the xml file.

Have you looked at what the "copyall" data mode actually does with the data? I assume it just passes the fields as is to the mediator?

> Regarding "presaero.clim_2000" and "presndep.clim_2000" data streams, I think we need aerosol and lnd input files.

Are you saying it's not possible for us to set up a configuration that doesn't have these inputs?

ezhilsabareesh8:
> Have you looked at what the "copyall" data mode actually does with the data? I assume it just passes the fields as is to the mediator?


I have reviewed the `copyall` data mode and can confirm that it passes the runoff stream variables `friver` and `licalvf` as-is to the mediator.

> Are you saying it's not possible for us to set up a configuration that doesn't have these inputs?

I was able to run the configuration without the aerosol and lnd input data. However, it is not entirely clear whether the aerosol input files are mandatory or optional. Based on my understanding, if the aerosol input data is not included in the XML files, datm nuopc is automatically ignoring it. [Refer (atm_comp_nuopc.f90)](https://github.com/ESCOMP/CDEPS/blob/f027aa64285fb9ddad9be5c5837a6e6e279e6051/datm/atm_comp_nuopc.F90#L781)

dougiesquire:
> I was able to run the configuration without the aerosol and lnd input data. However, it is not entirely clear whether the aerosol input files are mandatory or optional. Based on my understanding, if the aerosol input data is not included in the XML files, datm nuopc is automatically ignoring it. [Refer (atm_comp_nuopc.f90)](https://github.com/ESCOMP/CDEPS/blob/f027aa64285fb9ddad9be5c5837a6e6e279e6051/datm/atm_comp_nuopc.F90#L781)

Great - that's what we'd want to replicate ACCESS-OM2 as closely as possible right?

ezhilsabareesh8:
Thanks @dougiesquire. I have removed the aerosol and lnd input stream files from `datm.streams.xml` and `config.yaml`

ezhilsabareesh8 (commented):
Thanks @dougiesquire. I have made all the necessary changes.

dougiesquire (approved):
Looks good to me, but let's give @aekiss a chance to have a look if he wants to.

aekiss:
Thanks both - I'm busy with meetings for the rest of today but hope to find some time tomorrow

dougiesquire:
A thought.

For the data components, we currently remap the input data (640 x 320) onto a data model grid (192 x 94) that is different than both the input data grid and the ice/ocean grid. I’m not sure why this configuration was originally set up this way, but is this really something we want to do?

I think simply setting the data model grid (mesh file) to be the same as the input data grid will mean no spatial remapping occurs in CDEPS - https://escomp.github.io/CDEPS/versions/master/html/introduction.html#design. Alternatively, we could set the data model grid to be the same as the ocean/ice grid so that remapping happens in CDEPS but (hopefully) not in CMEPS.

Thoughts @ezhilsabareesh8, @aekiss?

datm_in
```diff
@@ -1,5 +1,5 @@
 &datm_nml
```
dougiesquire, line 1: Should `flds_presaero` and `flds_presndep` be set to `.false.` now that we've removed those streams?

aekiss:
Good catch re. remapping @dougiesquire. We definitely don't want to coarsen the JRA55-do input data!

Does CDEPS do time interpolation on the input grid or the output grid, ie before or after regridding?
- If time interpolation happens on the input grid, then I guess it doesn't matter whether regridding happens in CDEPS or CMEPS.
- But if time interpolation happens on the output grid in CDEPS, the relative computational costs of time interpolation and regridding might determine whether it's more efficient to regrid in CDEPS (ie time-interp regridded high-spatial-res data) or CMEPS (ie regrid high-temporal-res data). This difference might become significant at high resolution, since the data volume increases quadratically in space and linearly in time as the resolution is increased (and timestep decreases due to CFL).

dougiesquire:
Time interpolation is done after spatial remapping (see [here](https://escomp.github.io/CDEPS/versions/master/html/introduction.html#design) and [here](https://escomp.github.io/CDEPS/versions/master/html/design_details.html#data-model-performance)) so it seems like spatial remapping in CDEPS might be the better approach?

aekiss:
Yep, your [2nd link](https://escomp.github.io/CDEPS/versions/master/html/design_details.html#data-model-performance) says that regridding is more expensive than time interpolation (and I guess would become more so at high spatial resolution), so I agree we should regrid to the MOM/CICE model resolution in CDEPS.

datm_in
ezhilsabareesh8: In the current configuration, the flags `flds_preasero` and `flds_presndep` are set to `.true.` in `datm_in`. These flags are meant to control the inclusion of aerosol and lnd streams in the field list. However, even after removing the corresponding aerosol streams from the `datm.streams.xml` and `config.yaml` files, the code does not return any errors as expected.

It is evident that when `flds_presaero = .true.`, the aerosol streams are added to the field list in the source code [here](https://github.com/ESCOMP/CDEPS/blob/f027aa64285fb9ddad9be5c5837a6e6e279e6051/datm/datm_datamode_jra_mod.F90#L113-L118) and advertised [here](https://github.com/ESCOMP/CDEPS/blob/f027aa64285fb9ddad9be5c5837a6e6e279e6051/datm/datm_datamode_jra_mod.F90#L132)

However, even after removing the aerosol streams from the data streams configuration files, the mediator log shows that aerosol fields are still being advertised, which is unexpected behavior. 

``` 
(Advertise Fields):Fr_atm: Faxa_bcph 
(Advertise Fields):Fr_atm: Faxa_ocph
(Advertise Fields):Fr_atm: Faxa_dstwet
(Advertise Fields):Fr_atm: Faxa_dstdry 
```

I am checking the reason for this behaviour, since the removal of aerosol streams in the datm xml file with the flag `flds_presaero = .true` should return some error.

dougiesquire: Great - thanks for looking into this - I think it's important that we understand why we don't get the behaviour we expect

dougiesquire:
> Yep, your [2nd link](https://escomp.github.io/CDEPS/versions/master/html/design_details.html#data-model-performance) says that regridding is more expensive than time interpolation (and I guess would become more so at high spatial resolution), so I agree we should regrid to the MOM/CICE model resolution in CDEPS.

One wrinkle is that I can't see a way to pass precomputed remapping weights to CDEPS like you can for CMEPS remapping, but this should be something we could add.

datm_in
ezhilsabareesh8: Thanks @dougiesquire. Upon investigation into the code and the mediator log, it has been observed that the aerosol streams, specifically `Faxa_bcph, Faxa_ocph, Faxa_dstwet, and Faxa_dstdry`, are indeed added to the field list and advertised when the `flds_presaero` flag is set to `.true.`. Similarly, the land streams are also part of the field list and advertised when `flds_presndep` is set to `.true.`.

However, despite being included in the field list and advertised, there seems to be no utilization or call for these aerosol and land streams by other components. The mediator log indicates that these stream variables are not mapped into other components, which explains why there are no errors when we set `flds_presaero = .false.` and `flds_presndep = .false.` in the `datm_in` file.

```
 mapping atm->ocn Sa_u via patch_uv3d with one normalization
 mapping atm->ocn Sa_v via patch_uv3d with one normalization
 mapping atm->ocn Sa_z via bilnr with one normalization
 mapping atm->ocn Sa_tbot via bilnr with one normalization
 mapping atm->ocn Sa_pbot via bilnr with one normalization
 mapping atm->ocn Sa_shum via bilnr with one normalization
 mapping atm->ocn Sa_ptem via bilnr with one normalization
 mapping atm->ocn Sa_dens via bilnr with one normalization
 mapping atm->ocn Faxa_swnet via consf with one normalization
 mapping atm->ocn Faxa_rainc via consf with one normalization
 mapping atm->ocn Faxa_rainl via consf with one normalization
 mapping atm->ocn Faxa_snowc via consf with one normalization
 mapping atm->ocn Faxa_snowl via consf with one normalization
 mapping atm->ocn Faxa_lwdn via consf with one normalization
 mapping atm->ocn Faxa_swndr via consf with one normalization
 mapping atm->ocn Faxa_swvdr via consf with one normalization
 mapping atm->ocn Faxa_swndf via consf with one normalization
 mapping atm->ocn Faxa_swvdf via consf with one normalization
 mapping atm->ocn Sa_pslv via bilnr with one normalization
```


To clarify the configuration and avoid any confusion, I am setting `flds_presaero = .false.` and `flds_presndep = .false.` in the `datm_in` file to ensure that these aerosol and land streams are not included in the field list and are not advertised.

dougiesquire: Interesting. Is this the case even when streams are provided for `presaero` and `presndep`?

ezhilsabareesh8: Yes, that's correct

dougiesquire: Just checking I understand what you're saying: it's currently _not possible_ to use `presaero` and `presndep` forcing? Worse, it's possible to set things up so that it looks like `presaero` and `presndep` forcing is being applied, but the data is never actually being used?

ezhilsabareesh8: Mediator log reveals aerosol streamfiles are being mapped to the ice component, regardless of the presence of input aerosol files. 

```
 mapping atm->ice Faxa_bcph via consf with one normalization
 mapping atm->ice Faxa_dstwet via consf with one normalization
 mapping atm->ice Faxa_dstdry via consf with one normalization
```
This is unexpected behaviour,  since the code should return some error if there is no input files for the aerosol streams. For clarity, `flds_presaero` and `flds_presndep` are set to `.false.` in the current configuration and also `tr_aero` in `ice_in` need to be set to `.false.`.

micaeljtoliveira:
@aekiss @dougiesquire @ezhilsabareesh8 I've rebased this branch unto the new `1deg_jra55do_ryf` branch and changed the target of the PR.

It would be good if you had again a look at the changes to see if I didn't make a mistake when resolving conflicts.

dougiesquire (approved):
Looks good to me. I ran this for 1 day without issue.

PR #12, opened 2023-10-12
Restructure input directory

### LINKED ISSUE #10: Restructure input directories
See COSIMA/ACCESS-OM3#64 for details and motivation.

micaeljtoliveira:
Closes #10

### DIFF
--- a/manifests/input.yaml
+++ b/manifests/input.yaml
@@ -67,29 +67,29 @@ work/input/RYF.vas.1990_1991.nc:
     binhash: 17224eecb745b6ce72604b41201c1fe7
     md5: 88dc8c70338bbc8f5efc48ed0009a99f
 work/input/access-om2-1deg-ESMFmesh.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/access-om2-1deg-ESMFmesh.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/share/access-om2-1deg-ESMFmesh.nc
   hashes:
-    binhash: db1ebe07724c4b9350947423239e58f0
+    binhash: b76768044950ac2a165e3c2cd70e2f56
     md5: 670568bbb1f7f14c534b37ec895af118
 work/input/access-om2-1deg-nomask-ESMFmesh.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/access-om2-1deg-nomask-ESMFmesh.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/share/access-om2-1deg-nomask-ESMFmesh.nc
   hashes:
-    binhash: 465e2826fc0b9344eda8bcde953160a9
+    binhash: 25269214daf2808dbf536f434b707b71
     md5: d40d8512d683a089b6feb343d903ad71
 work/input/grid.nc:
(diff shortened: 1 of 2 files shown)


manifests/input.yaml
```diff
@@ -67,29 +67,29 @@ work/input/RYF.vas.1990_1991.nc:
...
 work/input/access-om2-1deg-nomask-ESMFmesh.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/access-om2-1deg-nomask-ESMFmesh.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/share/access-om2-1deg-nomask-ESMFmesh.nc
   hashes:
-    binhash: 465e2826fc0b9344eda8bcde953160a9
+    binhash: 25269214daf2808dbf536f434b707b71
     md5: d40d8512d683a089b6feb343d903ad71
 work/input/grid.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/grid.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/share/grid.nc
```
dougiesquire, line 80: This should be in the `cice` directory (you've probably already moved this for https://github.com/COSIMA/MOM6-CICE6-WW3/pull/12)

dougiesquire (commented):
One file to move that you might have already moved, depending on which PR you did first. Thanks!

dougiesquire (approved):

manifests/input.yaml
```diff
@@ -652,27 +652,27 @@ work/input/make_ryf/make_ryf.py:
...
 work/input/salt_sfc_restore.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/salt_sfc_restore.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/mom/salt_sfc_restore.nc
   hashes:
-    binhash: 5e68da4372238720dd46caae3b03787e
+    binhash: d1546a2b957a23e34f1154ad180df60d
     md5: b2bd35c44017597ba99b85fb61ed4d72
 work/input/topog.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/topog.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/share/topog.nc
```
aekiss, line 675: should be in `mom`, right?

aekiss (approved):
LGTM if 2 files are moved

manifests/input.yaml
dougiesquire: This file is actually used by both `mom` and `cice`

manifests/input.yaml
micaeljtoliveira: Good catch. Fixed.

manifests/input.yaml
micaeljtoliveira: Yes, the file is used in both, so I guess it will be needed for the CICE6-WW3 configuration, where we are not supposed to include the `mom` input directory.

aekiss: Ah ok, my mistake. It was in the mom dir for access-om2 but I guess CICE6 needs topography for knowing when to ground fast ice.

aekiss: so yes, in that case it would also be needed for CICE6-WW3

PR #17, opened 2023-10-27
Cice config

anton-seaice:
Copy OM2 config across from https://github.com/COSIMA/1deg_jra55_ryf/blob/master/ice/cice_in.nml

Noting 
- initial conditions are set per https://github.com/COSIMA/access-om3/issues/50 , and to move to using initial ice conditions based on SST and Latitude would need (very minor) code changes
- nprocs and the block sizes haven't been tuned (check in OM2 technical report)
- chunking in output files would need code changes (https://github.com/COSIMA/access-om3/issues/81)
- It looks like dalb_mlt was set in the namelist in OM2, but still hardcoded to the [default](https://github.com/COSIMA/cice5/blob/edcfa6f9c76ed05b63196ce4b5355fa5a8f4fe3a/source/ice_shortwave.F90#L712C10-L712C18)


anton-seaice:
@aekiss Can you review?

aekiss:
Great, thanks @anton-seaice I'll hopefully have a chance to look at this on Thursday.

aekiss:
Here's a [table](https://github.com/COSIMA/access-om3-doc/blob/main/tables/ice_in/MOM6-CICE6_cice_config_nml.md) comparing this PR to ACCESS-OM2 and the current MOM6-CICE6 settings. Differences are in bold.
[This table](https://github.com/COSIMA/access-om3-doc/blob/main/tables/ice_in/MOM6-CICE6_cice_config_nml_diff.md) shows only the differences.

anton-seaice:
> Here's a [table](https://github.com/COSIMA/access-om3-doc/blob/main/tables/ice_in/MOM6-CICE6_cice_config_nml.md) comparing this PR to ACCESS-OM2 and the current MOM6-CICE6 settings. Differences are in bold. [This table](https://github.com/COSIMA/access-om3-doc/blob/main/tables/ice_in/MOM6-CICE6_cice_config_nml_diff.md) shows only the differences.

Thanks - at some point we might want to figure out how to add a notes column to the table? (And ideally the default values too!)

dougiesquire:
> Here's a [table](https://github.com/COSIMA/access-om3-doc/blob/main/tables/ice_in/MOM6-CICE6_cice_config_nml.md) comparing this PR to ACCESS-OM2 and the current MOM6-CICE6 settings. Differences are in bold.
[This table](https://github.com/COSIMA/access-om3-doc/blob/main/tables/ice_in/MOM6-CICE6_cice_config_nml_diff.md) shows only the differences.

@anton-seaice, how come there are there are lots of variables in your namelist that are different to both ACCESS-OM2 and the current MOM6-CICE6?

anton-seaice:
> @anton-seaice, how come there are there are lots of variables in your namelist that are different to both ACCESS-OM2 and the current MOM6-CICE6?

Noting I deleted all the lines where we use the default value, other differences are:
- sometimes CICE5 compiler options are now CICE6 config options (e.g. nblyr)
- sometimes the directories/paths/filenames are controlled by NUOPC now instead of CICE
- I need to double check `albedo_type`

Did you have specific examples?

dougiesquire:
> Did you have specific examples?

Possibly most of what I was looking at is due to default stuff being removed. But yes, the `albedo_type` and also `write_ic` I suppose

anton-seaice:
There are a few items defined in both _ice_in_ and _nuopc.runconfig_ which is confusing at best:
- I deleted `nprocs` from _ice_in_ becuase the one in runconfig is the one that is used
- `ice_ncat`, `ice_nx`, `ice_ny` in _nuopc.runconfig_ appear to not be used (they are set also in _ice_in_ ) but I haven't deleted them.
- I didn't dig into the `ICE_modelio` section in _nuopc.runconfig_ but suspect most of those fields don't matter. 

I will turn off the writing initial conditions again, and set `albedo_type` to default in ice_in.

I thought / assumed all the output fields were not saved by default. By there is a whole list here: https://github.com/CICE-Consortium/CICE/blob/4450a3e8c64bc07d1173eb3e341cd8dea91d5068/cicecore/cicedyn/analysis/ice_history_shared.F90#L202 which are saved monthly by default, which I will need to work through to set to match the OM2 config.

anton-seaice:
I have gone through and fixed this to correct for the icefields which are on by default. Here are the CICE log extracts which list the output fields - based on the latest commit (https://github.com/COSIMA/MOM6-CICE6/pull/17/commits/76b2c60f37d53ac97edad05c4806e8a0c3801803).

[om2_ryf_example.log](https://github.com/COSIMA/MOM6-CICE6/files/13275814/om2_ryf_example.log)
[cice_config_extract.log](https://github.com/COSIMA/MOM6-CICE6/files/13275815/cice_config_extract.log)

 The only substantive difference is `fsensn_ai_m` which looks like if was off by mistake in OM2 (we were saving the closely related `flatn_ai_m`)

anton-seaice:
@aekiss Have you been able to have a look?

ice_in
```diff
@@ -1,450 +1,156 @@
...
   update_ocn_f = .true.
   ustar_min = 0.005
-  wave_spec_file = "unknown_wave_spec_file"
-  wave_spec_type = "none"
 /
 &domain_nml
-  add_mpi_barriers = .false.
   block_size_x = 20
   block_size_y = 16
-  distribution_type = "sectrobin"
```
aekiss, line 197: I'd prefer to keep these to make it explicit, as we'll be tuning them
```
  distribution_type = "sectrobin"
  distribution_wght = "blockall"
```

ice_in
```diff
@@ -1,450 +1,156 @@
 &setup_nml
-  bfbflag = "off"
```
aekiss, line 2: suggest leaving these in as a reminder of available debug options
```
  bfbflag = "off"
  conserv_check = .false.
```

ice_in
```diff
@@ -1,450 +1,156 @@
...
   maskhalo_bound = .true.
   maskhalo_dyn = .true.
   maskhalo_remap = .true.
-  max_blocks = 8
+  max_blocks = 10
   ns_boundary_type = "tripole"
   nx_global = 360
   ny_global = 300
-  processor_shape = "square-ice"
+  processor_shape = "slenderX2"
```
aekiss, line 75:
```suggestion
  processor_shape = "slenderX1"
```
to match access-om2

ice_in
```diff
@@ -1,450 +1,156 @@
...
-  dsdt_slow_mode = -1.5e-07
-  kitd = 1
-  ktherm = 2
-  phi_c_slow_mode = 0.05
-  phi_i_mushy = 0.85
-  rac_rapid_mode = 10
+  dsdt_slow_mode = -5e-08
 /
 &dynamics_nml
-  advection = "remap"
```
aekiss, line 96: suggest leaving this explicit as we may need to change it for C-grid
```
  advection = "remap"
```

ice_in
```diff
@@ -1,450 +1,156 @@
...
-  alphab = 20.0d0
-  arlx = 300.
-  brlx = 300.
-  cf = 17.0
-  coriolis = "latitude"
-  cstar = 20.0d0
-  e_plasticpot = 2.0
-  e_yieldcurve = 2.0
   k1 = 8.0d0
```
aekiss, line 36:
```suggestion
```
suggest using default landfast parameterisation here until we have a reason to change

ice_in
```diff
@@ -1,450 +1,156 @@
...
-  kstrength = 1
-  ktens = 0.0
-  ktransport = 1
-  mu_rdg = 4.0
-  ndte = 120
-  pstar = 2.75d4
-  revised_evp = .false.
-  seabed_stress = .false.
-  seabed_stress_method = "LKD"
-  ssh_stress = "coupled"
```
aekiss, line 120: should retain
```
  ssh_stress = "coupled"
```
since this isn't the default

ice_in
```diff
@@ -1,450 +1,156 @@
...
 /
 &forcing_nml
-  atmbndy = "default"
-  calc_strair = .true.
-  calc_tsfc = .true.
   emissivity = 0.95d0
-  fbot_xfer_type = "constant"
-  formdrag = .false.
   highfreq = .true.
-  ice_ref_salinity = 4.0d0
```
aekiss, line 180: use
```
  ice_ref_salinity = 5.0d0
```
to match `ice_salt_concentration` in MOM6 and access-om2

ice_in
```diff
@@ -1,450 +1,156 @@
...
-  l_mpond_fresh = .false.
-  natmiter = 5
-  nfreq = 25
-  restart_coszen = .false.
-  rotate_wind = .true.
-  saltflux_option = "prognostic"
-  tfrz_option = "mushy"
+  tfrz_option = "linear_salt"
   update_ocn_f = .true.
   ustar_min = 0.005
```
aekiss, line 63: Let's use
```suggestion
  ustar_min = 0.0005
```
for now to match access-om2. This is also used in https://github.com/CICE-Consortium/CICE/blob/b14cedfaed8b81500fc5422cfc44b6d80e5893ef/configuration/scripts/ice_in#L253. Also see https://bb.cgd.ucar.edu/cesm/threads/ustar_min-default-documentation.9059/

ice_in
anton-seaice: With the block size and nx/ny settings here, we have 18 blocks in the x direction and 19 in the y direction.

With slenderX1, this gives 19 blocks per processor but only 18 processors would be used. With slenderX2 we get 9/10 blocks per processor and 36 processors used.

Now that I write this, maybe we should change the block_size_x to 16 and block_size_y to 15 with slenderX2 as this gives still 10 blocks per processor but they are slightly smaller and all 48 processors are actually used.

ice_in
```diff
@@ -1,450 +1,156 @@
...
+  r_snw = 0.0
   sw_dtemp = 0.01d0
   sw_frac = 1.0d0
   sw_redist = .true.
 /
 &ponds_nml
   dpscale = 1.0e-3
-  frzpnd = "cesm"
-  hp1 = 0.01
-  hs0 = 0.0
```
aekiss, line 145: retain, as this isn't the default
```
  hs0 = 0.0
```

ice_in
```diff
@@ -1,450 +1,156 @@
...
-  albicei = 0.45
-  albicev = 0.75
-  albsnowi = 0.73
+  ahmax = 0.1
+  albicei = 0.44
+  albicev = 0.86
+  albsnowi = 0.7
   albsnowv = 0.98
-  dt_mlt = 1.50
+  dt_mlt = 1.0
```
aekiss, line 44:
```suggestion
  dt_mlt = 1.0
  dalb_mlt = -0.02
```
to match access-om2 since this isn't the default

ice_in
```diff
@@ -1,450 +1,156 @@
...
   kalg = 0.0
-  r_ice = 0.0
-  r_pnd = 0.0
-  r_snw = 1.25
-  rsnw_mlt = 1500.
-  shortwave = "dEdd"
+  r_snw = 0.0
   sw_dtemp = 0.01d0
   sw_frac = 1.0d0
   sw_redist = .true.
```
aekiss, line 49: these three seem to be for a new feature in cice6 - so should we set
```
sw_redist = .false.
```
for now to match access-om2?

aekiss: actually, that's the default, so all 3 could be deleted, right?

aekiss:
OK I've looked through everything, apart from diagnostics (which we'll adjust anyway) and BGC (which we can think about later, since CICE6 has significantly improved BGC capabilities).

I've noted just a few things to change and then it will be ready to merge.

aekiss:
I've also updated the [comparison table](https://github.com/COSIMA/access-om3-doc/blob/main/tables/ice_in/MOM6-CICE6_cice_config_nml_diff.md)

ice_in
anton-seaice: These might be related to the new snow scheme, so yes lets turn off.

ice_in
anton-seaice: This is only used if shortwave=dEdd (but we used shortwave default = ccsm3)

ice_in
anton-seaice: If would be nice if we can figure out the history / justification for these. It looks like in [earlier "AusCOM" definitions](https://github.com/COSIMA/cice5/blob/edcfa6f9c76ed05b63196ce4b5355fa5a8f4fe3a/source/ice_shortwave.F90#L83) dt_mlt meant 

_change in temp to give dalb_mlt albedo change" (which is hardcoded to 1 now)_

but now it [means](https://github.com/CICE-Consortium/Icepack/blob/f6ff8f7c4d4cb6feabe3651b13204cf43fc948e3/columnphysics/icepack_shortwave.F90#L3317C22-L3317C86) :

_change in temp for non-melt to melt snow grain radius change (C)_

And looks like its only used for _shortwave=dedd_ (we use default shortwave=ccsm3)

I guess dalb_mlt could have been set to reflect Antarctic vs Arctic differences? (although possibly surprising to not also set dalb_mltv dalb_mlti)

anton-seaice: It looks like dalb_mlt was added to the namelist for AusCOM and is not configurable by default. (https://github.com/COSIMA/cice5/blob/edcfa6f9c76ed05b63196ce4b5355fa5a8f4fe3a/source/ice_init.F90#L178)

ice_in
anton-seaice: Ah - I get this warning with it set to false:

```
 (input_data) WARNING: ktherm = 1 and sw_redist =  F
 (input_data) WARNING:   For consistency, set sw_redist = .true.
```

anton-seaice: It looks like it used to be on by default for BL99, so `sw_redist=.true.` would be correct to be consistent with OM2

ref: https://github.com/CICE-Consortium/Icepack/pull/326#pullrequestreview-444290515

anton-seaice:
Thanks @aekiss . I responded to all the comments I think

ice_in
aekiss: looks to me like AusCOM used `dalb_mlt`, but subsequently the same name was used in an update to cice to mean something else

ice_in
aekiss: ok it doesn't matter a lot at this stage as the core count will probably change

aekiss (approved):
thanks @anton-seaice, happy for you to merge

micaeljtoliveira:
> @aekiss - I don't have write access to this repo!

Should be fixed now.

PR #21, opened 2023-12-06
Set year_align=year_first in data streams for IAF configs

### LINKED ISSUE #20: year_align set incorrectly in IAF stream.xml files
The `year_align` parameter is set to 1 in the `*.stream.xml` files. This is suitable for the RYF configurations, but not for IAF. For IAF, `year_align` should be equal to `year_first` so that the simulation and forcing years are aligned. See [here](https://escomp.github.io/CDEPS/versions/master/html/streams.html#definitions-of-each-keys-used-in-stream-definition-file) for details.

dougiesquire:
See #20 for context

Closes #20

### DIFF
--- a/datm.streams.xml
+++ b/datm.streams.xml
@@ -7,7 +7,7 @@
     <dtlimit>1.5</dtlimit>
     <year_first>1958</year_first>
     <year_last>2018</year_last>
-    <year_align>1</year_align>
+    <year_align>1958</year_align>
     <vectors>null</vectors>
     <meshfile>./input/JRA55do-ESMFmesh.nc</meshfile>
     <lev_dimname>null</lev_dimname>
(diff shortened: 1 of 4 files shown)


anton-seaice:
Thoughts on setting `taxmode` to `limit` ? Which I think will reject dates outside the time range.

https://escomp.github.io/CDEPS/versions/master/html/streams.html#definitions-of-each-keys-used-in-stream-definition-file

dougiesquire:
> Thoughts on setting taxmode to limit ?

Could be a good idea, but @aekiss may have opinions (when he's back). Having IAF runs "auto-cycle" might be desirable?

dougiesquire:
> Thoughts on setting taxmode to limit ?

@anton-seaice shall we address this in a separate issue. Would you mind opening an issue for this?

PR #23, opened 2023-12-07
Cice config (#17)

anton-seaice:
* Update CICE config to match OM2 CICE Config

### DIFF
--- a/ice_in
+++ b/ice_in
@@ -1,450 +1,159 @@
 &setup_nml
-  bfbflag = "off"
+  bfbflag = "off" 
   conserv_check = .false.
-  days_per_year = 365
-  diagfreq = 24
-  dumpfreq = "x"
-  dumpfreq_n = 1
-  hist_avg = .true., .true., .true., .true., .true.
-  histfreq = "m", "x", "x", "x", "x"
-  histfreq_n = 1, 0, 0, 0, 0
-  history_file = "unknown"
-  history_format = "default"
-  history_precision = 4
+  debug_forcing = .true.
+  debug_model = .true.
+  diagfreq = 960
+  dumpfreq = "y"
+  dump_last = .true.
+  histfreq = "d", "m", "x", "x", "x"
   ice_ic = "./input/iced.1900-01-01-10800.nc"
-  latpnt = 90.0, -65.0
   lcdf64 = .true.
-  lonpnt = 0.0, -45.0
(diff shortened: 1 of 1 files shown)


aekiss:
LGTM, my only question is whether we should have `use_leap_years`=true for IAF. It was false for both IAF and RYF in ACCESS-OM2 for reasons I no longer recall, but [the calendar manager was updated in CICE6](https://cice-consortium-cice.readthedocs.io/en/main/user_guide/ug_implementation.html#time-manager) so we may need to revisit this. Or does NUOPC handle all the calendar stuff?

anton-seaice:
> LGTM, my only question is whether we should have `use_leap_years`=true for IAF. It was false for both IAF and RYF in ACCESS-OM2 for reasons I no longer recall, but [the calendar manager was updated in CICE6](https://cice-consortium-cice.readthedocs.io/en/main/user_guide/ug_implementation.html#time-manager) so we may need to revisit this. Or does NUOPC handle all the calendar stuff?

I'll investigate - I am pretty sure this config field is used, and we have leap years in the forcing, so you are probably right.

anton-seaice:
It looks like `ice_calendar` is set by NUOPC:

https://github.com/CICE-Consortium/CICE/blob/b14cedfaed8b81500fc5422cfc44b6d80e5893ef/cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90#L487

https://github.com/CICE-Consortium/CICE/blob/b14cedfaed8b81500fc5422cfc44b6d80e5893ef/cicecore/shared/ice_calendar.F90#L214

but `days_per_year` and `use_leap_years` are not.

Which means the model runs, but some of the attributes/comments are wrong in the output if its not set (this may be the only impact):

https://github.com/CICE-Consortium/CICE/blob/b14cedfaed8b81500fc5422cfc44b6d80e5893ef/cicecore/cicedyn/infrastructure/io/io_pio2/ice_history_write.F90#L227

https://github.com/CICE-Consortium/CICE/blob/b14cedfaed8b81500fc5422cfc44b6d80e5893ef/cicecore/cicedyn/infrastructure/io/io_pio2/ice_history_write.F90#L652

I doubt those matter, however maybe some analysis scripts would try and automatically parse that field. Either way, we should set it :)

anton-seaice:
Merged in #25

Issue #22, opened 2023-12-07
DATM & DROF xml config


anton-seaice:
Following on from #20 , two changes are proposed:

- In IAF configurations, for datm and drof, set `taxmode` to `limit` (doco: [here](https://github.com/COSIMA/MOM6-CICE6/blob/5797870f9bf82ff15d64ad4e4c97893c77bbcf64/drof.streams.xml#L4C5-L4C5)). This means that if the model date is set outside the range of provided forcing data, the date would be rejected. (Currently it is cycled around when outside the range). (https://escomp.github.io/CDEPS/versions/master/html/streams.html#definitions-of-each-keys-used-in-stream-definition-file)
- In RYF configurations, set `year_first`, `year_last` and `year_align` to be 1990 to be consistent with the input file. As data cycles around in the current setting, this does not make any functional change, but may be easier to interpret for future users. (The model year is set in the `nuopc.runconfig`, and we would leave it set to year 0 to be clear its not a real year).

dougiesquire:
> In RYF configurations, set year_first, year_last and year_align to be 1990

I think `year_first` and `year_last` need to remain at 1900, as the time variable in the files has been reset to span 1900 even though the data is from 1990-1991. But we could change `year_align` to 1900 so they are all consistent.

aekiss:
I think it's quite a good safety feature to have `taxmode=limit` to avoid users inadvertently getting into [leap-year tangles like these](https://github.com/COSIMA/access-om2/issues/149) (which are much worse with a forcing cycle not divisible by 4).

anton-seaice:
I think ill take that as consensus :)

Actions: 

- [ ] `generate_xml_datm.py` could be edited to change the comments re: year_align
- [ ] `generate_xml_datm.py` change taxmode to 'limit'
- [ ] generate new `datm.streams.xml` for IAF with taxmode='limit'
- [ ] change `datm.streams.xml` for RYF, to set year_align=1900

anton-seaice:
From @aekiss  on slack:

_We used 1900 in OM2 so I think we might as well stick with that. The first 4 mo of RYF9091 is from 1991 anyway.
Don't use 0000, since this may cause problems if any components use a strict definition of the Gregorian calendar as applying only from 15 October 1582 onwards, with a Julian calendar prior to that. https://github.com/COSIMA/access-om2/issues/117_


So I think we also change the start year in nuopc.runconfig to 1900

aekiss:
[...]

PR #28, opened 2024-01-10
limit valid years to 1958 to 2018

anton-seaice:
Turn off cycling around forcing data for iaf runs. #22


ezhilsabareesh8 (approved):
Looks good to me

PR #36, opened 2024-02-12
doc dir for wiki searches

aekiss:
Adds `doc` dir for wiki searches - see https://github.com/COSIMA/access-om3/issues/75

Includes an `update_doc.sh` to keep it up to date, and also update the changelog.


micaeljtoliveira (approved):
LGTM

PR #41, opened 2024-02-29
1deg_jra55do_ryf: change case name to "access-om3"

dougiesquire:
This PR changes the case name from "GMOM_JRA" to "access-om3". See https://github.com/COSIMA/access-om3/issues/106. The restart manifests accidentally added in https://github.com/COSIMA/MOM6-CICE6/pull/18 are also removed here.

The plan is to add a check to the ACCESS-OM3 Payu driver to check that the naming in the MOM `diag_table` is consistent with the case name set in `nuopc.runconfig`.

Once this PR is merged, I'll cherry pick into `1deg_jra55do_iaf`.


manifests/input.yaml
```diff
@@ -4,8 +4,8 @@ version: 1.0
 work/input/JRA55do-ESMFmesh.nc:
   fullpath: /g/data/ik11/inputs/access-om3/0.x.0/share/meshes/JRA55do-ESMFmesh.nc
   hashes:
-    binhash: 5195a7517bd0905add0f97ceecc0503f
-    md5: 154215afc3b742e3da9ac77e4eac6813
+    binhash: 1147443c3fe237a480af1b4700159e21
```
anton-seaice, line 7: Why has this file changed?

manifests/restart.yaml
anton-seaice, line 1: Can you please put this in a seperate commit to the rest of the change? (And then do a "Rebase and merge" when you merge it).

manifests/input.yaml
dougiesquire: It changed when the mesh generation script got moved to the `om3-scripts` repo (all that changed is the metadata about how the file was generated - the contents are the same)

anton-seaice: Maybe that needs its own commit too :)

manifests/restart.yaml
dougiesquire: Done

anton-seaice (approved):
I ran this as written and with a new build from [access-om3](https://github.com/anton-seaice/access-om3/commit/4b4c1892c3444f67f7bd98045922c63f0953e5f8) and it works. All the history output files (except [mom_rho2](https://github.com/COSIMA/MOM6-CICE6/issues/40)) and restart files are there with the `access-om3` name. Its possible to write cmeps history output too.

PR #42, opened 2024-03-04
1deg_jra55do_iaf: change case name to "access-om3"

dougiesquire:
Cherry-picked commits from #41. See https://github.com/COSIMA/access-om3/issues/106


anton-seaice (approved):
I check this out and it runs with same comments as RYF (https://github.com/COSIMA/MOM6-CICE6/pull/41#pullrequestreview-1913130824)

PR #47, opened 2024-03-27
Configure WOMBATlite in `1deg_jra55do_ryf_wombatlite`

dougiesquire:
This PR sets up the `1deg_jra55do_ryf_wombatlite` configuration to run with the new generic tracer implementation of WOMBATlite.


minghangli-uni:
I tried to run it but ended up with errors,

```
1619178 WARNING from PE     1: drtsafe: root not bracketed
1619179
1619180
1619181 WARNING from PE     1: drtsafe: root not bracketed
1619182
1619183
1619184 FATAL from PE    26: NaN in input field of reproducing_EFP_sum(_2d).
1619185
1619186
1619187 FATAL from PE    10: NaN in input field of reproducing_EFP_sum(_2d).
```

Also I've seen many warning messages, such as 
```
     27 WARNING from PE     0: Previous msg:Z_INIT_ALE_REMAPPING = True     !   [Boolean] default = False
     28
     29
     30 WARNING from PE     0: New message :Z_INIT_ALE_REMAPPING = True     !   [Boolean] default = True
```
I am not sure about it. Are we expecting these warnings for WOMBAT?

dougiesquire:
> Are we expecting these warnings for WOMBAT?

No, these are not expected. These warnings are introduced after @aekiss's changes in #94 to increase `DT_THERM=6*DT`. To run WOMBAT with `THERMO_SPANS_COUPLING` requires a configuration change and I haven't tested this yet.

**I think we should revert the changes from https://github.com/COSIMA/MOM6-CICE6/pull/94**. They haven't been tested. Also, why was the coupling timestep halved? @aekiss and @minghangli-uni are you okay for me to revert [this commit](https://github.com/COSIMA/MOM6-CICE6/commit/58e3b755e8e09664f4c8eedb40f428ce215a4bfb)?

minghangli-uni:
Thanks @dougiesquire. Halving the coupling timestep is to avoid the unused buoyancy fluxes crash. See https://github.com/COSIMA/access-om3/issues/195 and https://github.com/NCAR/MOM6/issues/290

dougiesquire:
> Halving the coupling timestep is to avoid the unused buoyancy fluxes crash.

Thanks @minghangli-uni. I think it would be better to understand and resolve the issue if we can. Halving the coupling time step will unnecessary additional overhead.

minghangli-uni:
I will revisit this issue https://github.com/COSIMA/access-om3/issues/195.

minghangli-uni (approved):
Thanks @dougiesquire. I successfully ran it for a month and am comfortable with merging it. But one thing to note that without spanning dt_therm, there are still some warnings, such as this one:

```
225 WARNING from PE     0: Previous msg:Z_INIT_ALE_REMAPPING = True     !   [Boolean] default = False
226
227
228 WARNING from PE     0: New message :Z_INIT_ALE_REMAPPING = True     !   [Boolean] default = True
```

dougiesquire:
> Thanks @dougiesquire. I successfully ran it for a month and am comfortable with merging it. But one thing to note that without spanning dt_therm, there are still some warnings, such as this one:
>
> ```
> 225 WARNING from PE     0: Previous msg:Z_INIT_ALE_REMAPPING = True     !   [Boolean] default = False
> 226
> 227
> 228 WARNING from PE     0: New message :Z_INIT_ALE_REMAPPING = True     !   [Boolean] default = True> ```
> ```

Sorry yes I forgot to mention - this is because the default for this parameter is different in different modules. These should be resolved in the future but I think the warnings are okay for now.

PR #48, opened 2024-04-02
Upgrade 1-Degree to 025-Degree RYF Configuration

ezhilsabareesh8:
This pull request addresses the upgrade of the 1-Degree RYF configuration to a higher resolution of 0.25-Degree RYF configuration. 

Changes Implemented:
1. A new CICE grid has been generated from the MOM super grid to resolve the longitude mismatch issue in these PRs [1](https://github.com/COSIMA/om3-scripts/pull/6) and [2](https://github.com/COSIMA/om3-scripts/pull/8)

2. The CFL violation error caused by bad departure points has been fixed by adjusting the time step of CICE and MOM, similar to OM2 quarter-degree configurations. Note that only the time step has been changed and `ndtd` is not changed to avoid truncation errors in MOM6 [refer here](https://github.com/COSIMA/MOM6-CICE6/commit/96eebce8d2e7256cae053a94feac602ad66c9770).

3. Fixed Restart File Issue: Multiple restart files were being generated by MOM6 due to netCDF file limits. This issue has been addressed and fixed in this [PR](https://github.com/payu-org/payu/pull/432).

4. Updated `datm_in`, `drof_in`, and `nuopc.runconfig` files with a nomask ESMF mesh file for atmosphere and runoff components to resolve NaN.

5. Used a default ice initial condition for cold start, which will be updated with the initial conditions generated by running this configuration for a longer period.

6. All input files for this configuration have been updated with OM2 input files [refer here](https://github.com/COSIMA/access-om3/issues/124#issuecomment-2021659966).

7. This 025 degree configuration has been tested and executed for two years without encountering any errors.

Note that the MOM6 parameters listed [here](https://forum.access-hive.org.au/t/namelist-configuration-discussion-meeting/1917/8) needs to be updated and the block size in `ice_in` needs to be optimised for performance.

### DIFF
--- a/MOM_input
+++ b/MOM_input
@@ -12,7 +12,7 @@ THICKNESSDIFFUSE = True         !   [Boolean] default = False
 THICKNESSDIFFUSE_FIRST = True   !   [Boolean] default = False
                                 ! If true, do thickness diffusion or interface height smoothing before dynamics.
                                 ! This is only used if THICKNESSDIFFUSE or APPLY_INTERFACE_FILTER is true.
-DT = 1800.0                     !   [s]
+DT = 1350.0                     !   [s]
                                 ! The (baroclinic) dynamics time step.  The time-step that is actually used will
                                 ! be an integer fraction of the forcing time-step (DT_FORCING in ocean-only mode
                                 ! or the coupling timestep in coupled mode.)
@@ -63,10 +63,10 @@ SAVE_INITIAL_CONDS = True       !   [Boolean] default = False
 TRIPOLAR_N = True               !   [Boolean] default = False
                                 ! Use tripolar connectivity at the northern edge of the domain.  With
                                 ! TRIPOLAR_N, NIGLOBAL must be even.
-NIGLOBAL = 360                  !
+NIGLOBAL = 1440                 !
                                 ! The total number of thickness grid points in the x-direction in the physical
                                 ! domain. With STATIC_MEMORY_ this is set in MOM_memory.h at compile time.
-NJGLOBAL = 300                  !
+NJGLOBAL = 1080                 !
                                 ! The total number of thickness grid points in the y-direction in the physical
                                 ! domain. With STATIC_MEMORY_ this is set in MOM_memory.h at compile time.
 
@@ -119,7 +119,7 @@ MINIMUM_DEPTH = 0.5             !   [m] default = 0.0
                                 ! MASKING_DEPTH are rounded to MINIMUM_DEPTH.
 MAXIMUM_DEPTH = 6000.0          !   [m]
                                 ! The maximum depth of the ocean.
-CHANNEL_CONFIG = "global_1deg"  ! default = "none"
+CHANNEL_CONFIG = none           ! default = "none"
                                 ! A parameter that determines which set of channels are
                                 ! restricted to specific  widths.  Options are:
                                 !     none - All channels have the grid width.
@@ -588,7 +588,7 @@ HBD_LINEAR_TRANSITION = True    !   [Boolean] default = False
 ! === module ocean_stochastics_init ===
 
 ! === module ocean_model_init ===
-RESTART_CONTROL = 3             ! default = 1
+RESTART_CONTROL = 1             ! default = 1
                                 ! An integer whose bits encode which restart files are written. Add 2 (bit 1)
                                 ! for a time-stamped file, and odd (bit 0) for a non-time-stamped file.  A
                                 ! restart file will be saved at the end of the run segment for any non-negative
--- a/MOM_input
+++ b/MOM_input
@@ -588,11 +588,6 @@ HBD_LINEAR_TRANSITION = True    !   [Boolean] default = False
 ! === module ocean_stochastics_init ===
 
 ! === module ocean_model_init ===
-RESTART_CONTROL = 1             ! default = 1
-                                ! An integer whose bits encode which restart files are written. Add 2 (bit 1)
-                                ! for a time-stamped file, and odd (bit 0) for a non-time-stamped file.  A
-                                ! restart file will be saved at the end of the run segment for any non-negative
-                                ! value.
 OCEAN_SURFACE_STAGGER = "A"     ! default = "C"
                                 ! A case-insensitive character string to indicate the staggering of the surface
                                 ! velocity field that is returned to the coupler.  Valid values include 'A',
--- a/MOM_input
+++ b/MOM_input
@@ -16,7 +16,7 @@ DT = 1350.0                     !   [s]
                                 ! The (baroclinic) dynamics time step.  The time-step that is actually used will
                                 ! be an integer fraction of the forcing time-step (DT_FORCING in ocean-only mode
                                 ! or the coupling timestep in coupled mode.)
-DT_THERM = 3600.0               !   [s] default = 1800.0
+DT_THERM = 1350.0               !   [s] default = 1800.0
                                 ! The thermodynamic and tracer advection time step. Ideally DT_THERM should be
                                 ! an integer multiple of DT and less than the forcing or coupling time-step,
                                 ! unless THERMO_SPANS_COUPLING is true, in which case DT_THERM can be an integer
--- a/config.yaml
+++ b/config.yaml
@@ -9,20 +9,20 @@
 # shortpath: /scratch/v45
 
 queue: normal
-ncpus: 48
+ncpus: 240
 jobfs: 10GB
-mem: 192GB
+mem: 960GB
 
-walltime: 01:00:00
-jobname: 1deg_jra55do_ryf
+walltime: 48:00:00
+jobname: 025deg_jra55do_ryf
 
 model: access-om3
 
-exe: /g/data/ik11/spack/0.20.1/opt/linux-rocky8-cascadelake/intel-2021.6.0/access-om3-main-lxbeca7/bin/access-om3-MOM6-CICE6
+exe:  /g/data/ik11/spack/0.20.1/opt/linux-rocky8-cascadelake/intel-2021.6.0/access-om3-main-lxbeca7/bin/access-om3-MOM6-CICE6
 input: 
-    - /g/data/ik11/inputs/access-om3/0.x.0/1deg/share # shared grids and topography
-    - /g/data/ik11/inputs/access-om3/0.x.0/1deg/mom # grids, ICs etc
-    - /g/data/ik11/inputs/access-om3/0.x.0/1deg/cice # grids, ICs etc
+    - /g/data/ik11/inputs/access-om3/0.x.0/025deg/share # shared grids and topography
+    - /g/data/ik11/inputs/access-om3/0.x.0/025deg/mom # grids, ICs etc
+    - /g/data/ik11/inputs/access-om3/0.x.0/025deg/cice # grids, ICs etc
     - /g/data/ik11/inputs/access-om3/0.x.0/share/meshes/JRA55do-ESMFmesh.nc # mesh for JRA55-do stream
     - /g/data/ik11/inputs/JRA-55/RYF/v1-4 # datm and drof JRA55-do streams data
  
--- a/config.yaml
+++ b/config.yaml
@@ -18,7 +18,7 @@ jobname: 025deg_jra55do_ryf
 
 model: access-om3
 
-exe:  /g/data/ik11/spack/0.20.1/opt/linux-rocky8-cascadelake/intel-2021.6.0/access-om3-main-lxbeca7/bin/access-om3-MOM6-CICE6
+exe: /g/data/ik11/spack/0.20.1/opt/linux-rocky8-cascadelake/intel-2021.6.0/access-om3-main-lxbeca7/bin/access-om3-MOM6-CICE6
 input: 
     - /g/data/ik11/inputs/access-om3/0.x.0/025deg/share # shared grids and topography
     - /g/data/ik11/inputs/access-om3/0.x.0/025deg/mom # grids, ICs etc
--- a/manifests/input.yaml
+++ b/manifests/input.yaml
@@ -66,31 +66,26 @@ work/input/RYF.vas.1990_1991.nc:
   hashes:
     binhash: 17224eecb745b6ce72604b41201c1fe7
     md5: 88dc8c70338bbc8f5efc48ed0009a99f
-work/input/access-om2-1deg-ESMFmesh.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/share/access-om2-1deg-ESMFmesh.nc
+work/input/access-om2-025deg-ESMFmesh.nc:
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/share/access-om2-025deg-ESMFmesh.nc
   hashes:
-    binhash: c741e080783ac65f2a2e67a9cfdb7645
-    md5: 44e508ac57b244fc357faf5b12933bed
-work/input/access-om2-1deg-nomask-ESMFmesh.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/share/access-om2-1deg-nomask-ESMFmesh.nc
+    binhash: 5f29f55788f9fc9732c3bd713847d27e
+    md5: c006851eea2a64e881c41f3e9426e4ac
+work/input/access-om2-025deg-nomask-ESMFmesh.nc:
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/share/access-om2-025deg-nomask-ESMFmesh.nc
   hashes:
-    binhash: 491167fde697289421ff25be582d001d
-    md5: 9b7120a42b5cb587492e7c31791eb549
+    binhash: 7ae247d56106c51c1f7a58deda307210
+    md5: db5407804d759435c6846d2d2c661a6e
 work/input/grid.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/cice/grid.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/cice/grid.nc
   hashes:
-    binhash: 0952424dafa216e7c5d0f29ef96a7cb8
-    md5: 1213e346055ee073fe33dc12578d99c6
-work/input/iced.1900-01-01-10800.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/cice/iced.1900-01-01-10800.nc
-  hashes:
-    binhash: 2dfb4fa5231df7e352155c266adb29c5
-    md5: 87c012d60c58c65bb56caa98779e5e51
+    binhash: f1239d0326cebb637d0eabc6834e00c9
+    md5: 6946bd45cb5cbf932f869b39eb68bee2
 work/input/kmt.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/cice/kmt.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/cice/kmt.nc
   hashes:
-    binhash: cb7ff26d756687fedc897920818df02a
-    md5: 1f9806c646a620378e5257e480bc9df7
+    binhash: 52a74c8ec631ee045cf308a16af0d46c
+    md5: d87ca7229f1a579af6f3b308af16e3d1
 work/input/make_rhuss/.git/FETCH_HEAD:
   fullpath: /g/data/ik11/inputs/JRA-55/RYF/v1-4/make_rhuss/.git/FETCH_HEAD
   hashes:
@@ -652,27 +647,27 @@ work/input/make_ryf/make_ryf.py:
     binhash: 9fbb6a9f6bec50ccb2afafecae9a5ff7
     md5: c5ef2aeff9516ec76c80e76aef08d352
 work/input/ocean_hgrid.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/mom/ocean_hgrid.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/mom/ocean_hgrid.nc
   hashes:
-    binhash: 7f9cf6930cda20f2ba304eef45c2a97d
-    md5: 51f58be0f4ea6da2cb438a893f95c689
+    binhash: e0b2f3bb2d76badb35e2dfc07788738f
+    md5: 38b6f324ae16cc13c180699123ad85b5
 work/input/ocean_temp_salt.res.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/mom/ocean_temp_salt.res.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/mom/ocean_temp_salt.res.nc
   hashes:
-    binhash: 4812809ecd2abcc47351263cd2c66821
-    md5: c5f7e60b5427a4442f111adc65a2d067
+    binhash: 1c27f9fbe2c7755e4657ed14e5e14145
+    md5: 1797b25e6339595a320076080145fd5a
 work/input/ocean_vgrid.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/mom/ocean_vgrid.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/mom/ocean_vgrid.nc
   hashes:
-    binhash: 30af5058d31d9133770757144f6bd7cf
-    md5: a1e0b0b0adc363506ff17a362e83f64a
+    binhash: a10a26a7c361279ae0032afc8c83fadd
+    md5: e56e2936499d3ad53429d5940821118d
 work/input/salt_sfc_restore.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/mom/salt_sfc_restore.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/mom/salt_sfc_restore.nc
   hashes:
-    binhash: d1546a2b957a23e34f1154ad180df60d
-    md5: b2bd35c44017597ba99b85fb61ed4d72
+    binhash: abaea44efc3b0b31ab1a8eb9f46da744
+    md5: 434b4bdb323d17e9714d02519627fea2
 work/input/topog.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/share/topog.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/share/topog.nc
   hashes:
-    binhash: 049d8b297efe931a1c1d3f40af23afea
-    md5: 4e13d88001b646f3cf4f1c3b8db59f91
+    binhash: a46aeba8e75a3b4c8cd599e17c4dd0eb
+    md5: fa71dfdf4a91198d651d657bcf9aadb6
--- a/nuopc.runconfig
+++ b/nuopc.runconfig
@@ -284,9 +284,9 @@ CLOCK_attributes::
      glc_avg_period = yearly
      glc_cpl_dt = 86400
      history_ymd = -999
-     ice_cpl_dt = 1350
+     ice_cpl_dt = 3600
      lnd_cpl_dt = 3600
-     ocn_cpl_dt = 1350
+     ocn_cpl_dt = 3600
      restart_n = 1
      restart_option = nyears
      restart_ymd = -999
--- a/nuopc.runconfig
+++ b/nuopc.runconfig
@@ -300,7 +300,7 @@ CLOCK_attributes::
      tprof_n = -999
      tprof_option = never
      tprof_ymd = -999
-     wav_cpl_dt = 1800
+     wav_cpl_dt = 3600
 ::
 
 ATM_attributes::
--- a/nuopc.runconfig
+++ b/nuopc.runconfig
@@ -377,9 +377,9 @@ ICE_modelio::
      logfile = ice.log
      pio_async_interface = .false.
      pio_netcdf_format = nothing
-     pio_numiotasks = 1
+     pio_numiotasks = 5
      pio_rearranger = 1
-     pio_root = 0
+     pio_root = 1
      pio_stride = 48
      pio_typename = netcdf4p
 ::
--- a/nuopc.runconfig
+++ b/nuopc.runconfig
@@ -27,11 +27,11 @@ DRIVER_attributes::
 ::
 
 PELAYOUT_attributes::
-     atm_ntasks = 48
+     atm_ntasks = 240
      atm_nthreads = 1
      atm_pestride = 1
      atm_rootpe = 0
-     cpl_ntasks = 48
+     cpl_ntasks = 240
      cpl_nthreads = 1
      cpl_pestride = 1
      cpl_rootpe = 0
@@ -40,31 +40,31 @@ PELAYOUT_attributes::
      esp_nthreads = 1
      esp_pestride = 1
      esp_rootpe = 0
-     glc_ntasks = 48
+     glc_ntasks = 240
      glc_nthreads = 1
      glc_pestride = 1
      glc_rootpe = 0
-     ice_ntasks = 48
+     ice_ntasks = 240
      ice_nthreads = 1
      ice_pestride = 1
      ice_rootpe = 0
-     lnd_ntasks = 48
+     lnd_ntasks = 240
      lnd_nthreads = 1
      lnd_pestride = 1
      lnd_rootpe = 0
      ninst = 1
-     ocn_ntasks = 48
+     ocn_ntasks = 240
      ocn_nthreads = 1
      ocn_pestride = 1
      ocn_rootpe = 0
      pio_asyncio_ntasks = 0
      pio_asyncio_rootpe = 1
      pio_asyncio_stride = 0
-     rof_ntasks = 48
+     rof_ntasks = 240
      rof_nthreads = 1
      rof_pestride = 1
      rof_rootpe = 0
-     wav_ntasks = 48
+     wav_ntasks = 240
      wav_nthreads = 1
      wav_pestride = 1
      wav_rootpe = 0
@@ -116,12 +116,12 @@ ALLCOMP_attributes::
      hostname = gadi
      ice_ncat = 5
      mediator_present = true
-     mesh_atm = ./input/access-om2-1deg-nomask-ESMFmesh.nc
+     mesh_atm = ./input/access-om2-025deg-nomask-ESMFmesh.nc
      mesh_glc = UNSET
-     mesh_ice = ./input/access-om2-1deg-ESMFmesh.nc
+     mesh_ice = ./input/access-om2-025deg-ESMFmesh.nc
      mesh_lnd = UNSET
-     mesh_mask = ./input/access-om2-1deg-ESMFmesh.nc
-     mesh_ocn = ./input/access-om2-1deg-ESMFmesh.nc
+     mesh_mask = ./input/access-om2-025deg-ESMFmesh.nc
+     mesh_ocn = ./input/access-om2-025deg-ESMFmesh.nc
      model_version = unknown
      ocn2glc_coupling = .false.
      ocn2glc_levels = 1:10:19:26:30:33:35
@@ -147,8 +147,8 @@ MED_attributes::
      atm2lnd_map = unset
      atm2ocn_map = unset
      atm2wav_map = unset
-     atm_nx = 360
-     atm_ny = 300
+     atm_nx = 1440
+     atm_ny = 1080
      budget_ann = 1
      budget_daily = 0
      budget_inst = 0
@@ -253,8 +253,8 @@ MED_attributes::
      history_option_wav_inst = never
      ice2atm_map = unset
      ice2wav_smapname = unset
-     ice_nx = 360
-     ice_ny = 300
+     ice_nx = 1440
+     ice_ny = 1080
      info_debug = 1
      lnd2atm_map = unset
      lnd2rof_map = unset
@@ -263,15 +263,15 @@ MED_attributes::
      mapuv_with_cart3d = .true.
      ocn2atm_map = unset
      ocn2wav_smapname = unset
-     ocn_nx = 360
-     ocn_ny = 300
+     ocn_nx = 1440
+     ocn_ny = 1080
      ocn_surface_flux_scheme = 0
      rof2lnd_map = unset
      rof2ocn_fmapname = unset
      rof2ocn_ice_rmapname = unset
      rof2ocn_liq_rmapname = unset
-     rof_nx = 360
-     rof_ny = 300
+     rof_nx = 1440
+     rof_ny = 1080
      wav2ocn_smapname = unset
      wav_nx = 0
      wav_ny = 0
@@ -284,23 +284,23 @@ CLOCK_attributes::
      glc_avg_period = yearly
      glc_cpl_dt = 86400
      history_ymd = -999
-     ice_cpl_dt = 3600
+     ice_cpl_dt = 1350
      lnd_cpl_dt = 3600
-     ocn_cpl_dt = 3600
+     ocn_cpl_dt = 1350
      restart_n = 1
-     restart_option = nmonths
+     restart_option = nyears
      restart_ymd = -999
      rof_cpl_dt = 3600
      start_tod = 0
      start_ymd = 19000101
      stop_n = 1
-     stop_option = nmonths
+     stop_option = nyears
      stop_tod = 0
      stop_ymd = -999
      tprof_n = -999
      tprof_option = never
      tprof_ymd = -999
-     wav_cpl_dt = 3600
+     wav_cpl_dt = 1800
 ::
 
 ATM_attributes::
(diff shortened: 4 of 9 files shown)


manifests/input.yaml
```diff
@@ -66,31 +66,41 @@ work/input/RYF.vas.1990_1991.nc:
...
+  hashes:
+    binhash: 62206020c8390971f3f70ff6d1e2eab5
+    md5: 9f5055fb3d36eecd964f5f1c98dd6a80
+work/input/access-om2-nomask-025deg-ESMFmesh.nc:
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/share/access-om2-nomask-025deg-ESMFmesh.nc
+  hashes:
+    binhash: dbaa2ffa55d29117d433ea1e7cd3afbd
+    md5: 3ca48372532f9b9b7890936d600ac509
+work/input/chl.nc:
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/mom/chl.nc
```
anton-seaice, line 80: Do we need this file?

manifests/input.yaml
```diff
@@ -66,31 +66,41 @@ work/input/RYF.vas.1990_1991.nc:
...
+  hashes:
+    binhash: 3079228e9abaab3193b52b3109bd068b
+    md5: 99b68aec777292d746e4d93c2b976189
+work/input/global_gx1v5_grid.nc:
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/cice/global_gx1v5_grid.nc
```
anton-seaice, line 85: Please delete this from the input folder, its not used.

manifests/input.yaml
```diff
@@ -66,31 +66,41 @@ work/input/RYF.vas.1990_1991.nc:
...
+  hashes:
+    binhash: 341d37e811ef29789441c37d6ea8e5f1
+    md5: bacabd01f14a75e4e68d136c3b7b9c54
 work/input/grid.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/cice/grid.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/share/grid.nc
```
anton-seaice, line 90: I think this should be in the cice folder not the share folder ?

manifests/input.yaml
```diff
@@ -66,31 +66,41 @@ work/input/RYF.vas.1990_1991.nc:
...
   hashes:
-    binhash: 0952424dafa216e7c5d0f29ef96a7cb8
-    md5: 1213e346055ee073fe33dc12578d99c6
-work/input/iced.1900-01-01-10800.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/cice/iced.1900-01-01-10800.nc
+    binhash: 0962dcfdeb454b2b0a192bba573f2c2d
+    md5: 1967bc8dd2f8041f59bb4d98f856e3ed
+work/input/i2o.nc:
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/cice/i2o.nc
```
anton-seaice, line 95: Also pls delete this one

manifests/input.yaml
```diff
@@ -651,28 +661,83 @@ work/input/make_ryf/make_ryf.py:
   hashes:
     binhash: 9fbb6a9f6bec50ccb2afafecae9a5ff7
     md5: c5ef2aeff9516ec76c80e76aef08d352
+work/input/mask_table.464.48x40:
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/mom/mask_table.464.48x40
+  hashes:
+    binhash: 339b390c248283a1bc8113dadf3d4242
+    md5: 7a2fcdfbb9b331161c7d8874e2b7e6f9
+work/input/meshes/JRA55do-ESMFmesh.nc:
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/share/meshes/JRA55do-ESMFmesh.nc
```
anton-seaice, line 670: I dont think this file is specific to 0.25 degree right? Its the same file as 1degree? Can we put it in a common folder?

manifests/input.yaml
```diff
@@ -651,28 +661,83 @@ work/input/make_ryf/make_ryf.py:
...
+  hashes:
+    binhash: 3654d5a1e59bff0fe76104c7abaa0b10
+    md5: 1c8481b8c85f59b6cbfc2ed5bc24c77d
+work/input/monthly_sstsss.nc:
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/cice/monthly_sstsss.nc
```
anton-seaice, line 675: Also pls delete

manifests/input.yaml
```diff
@@ -651,28 +661,83 @@ work/input/make_ryf/make_ryf.py:
...
+  hashes:
+    binhash: 2f705ff1c81531a8da76174241fb0401
+    md5: 598c7aa8a08fe3f48a621f781dd26331
+work/input/o2i.nc:
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/cice/o2i.nc
```
anton-seaice, line 680: also pls delete

manifests/input.yaml
```diff
@@ -651,28 +661,83 @@ work/input/make_ryf/make_ryf.py:
...
+  hashes:
+    binhash: a5a48813d3a6bea89eb857850f08b7a0
+    md5: 5c695c6a91b4a2f64e653a5baa62c34a
 work/input/ocean_hgrid.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/mom/ocean_hgrid.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/share/ocean_hgrid.nc
```
anton-seaice, line 685:
```suggestion
  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/mom/ocean_hgrid.nc
```

manifests/input.yaml
```diff
@@ -651,28 +661,83 @@ work/input/make_ryf/make_ryf.py:
...
+  hashes:
+    binhash: ab2ac59eaebd7b46e1d749777f901f67
+    md5: 38b6f324ae16cc13c180699123ad85b5
+work/input/ocean_mask.nc:
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/share/ocean_mask.nc
```
anton-seaice, line 690:
```suggestion
  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/mom/ocean_mask.nc
```

metadata.yaml
anton-seaice, line 1: Are we adding this to every config? Is that the right thing to do? (@ Dougie?)

config.yaml
```diff
@@ -2,27 +2,27 @@
 
 # If submitting to a different project to your default, uncomment line below 
 # and change project code as appropriate; also set shortpath below
-project: x77
+project: tm70
```
anton-seaice, line 5: I guess don't change this until https://github.com/COSIMA/access-om3/issues/114 is implemented

config.yaml
```diff
@@ -2,27 +2,27 @@
...
 
 # Force payu to always find, and save, files in this scratch project directory
 # (you may need to add the corresponding PBS -l storage flag in sync_data.sh)
-shortpath: /scratch/v45
+shortpath: /scratch/tm70
```
anton-seaice, line 9: also revert this

anton-seaice (commented):
Well done Ezhil! Apologies for many questions!

Can you rebase this to use the latest RYF config changed please?

> 5\. Used a default ice initial condition for cold start, which will be updated with the initial conditions generated by running this configuration for a longer period.

For the 1deg we used initial conditions from 3 hour run of OM2, did we still want to generate those?

Shall we update `ice_in` to match the OM2 config now or in a later PR?

Does @minghangli-uni have suggestions for performance changes or is that also a later PR?

Shall we try and make a clean commit history to explain each change of somesort? (Or is this all one merge commit?)

manifests/input.yaml
ezhilsabareesh8: Update input and restart yaml files in this commit 058e5f7

manifests/input.yaml
ezhilsabareesh8: Update input and restart yaml files in this commit https://github.com/COSIMA/MOM6-CICE6/commit/058e5f7ee3ae4fd031bc1e9f87ae1e74ad263382

manifests/input.yaml
ezhilsabareesh8: Update input and restart yaml files in this commit https://github.com/COSIMA/MOM6-CICE6/commit/058e5f7ee3ae4fd031bc1e9f87ae1e74ad263382

manifests/input.yaml
ezhilsabareesh8: Update input and restart yaml files in this commit https://github.com/COSIMA/MOM6-CICE6/commit/058e5f7ee3ae4fd031bc1e9f87ae1e74ad263382

manifests/input.yaml
ezhilsabareesh8: Update input and restart yaml files in this commit https://github.com/COSIMA/MOM6-CICE6/commit/058e5f7ee3ae4fd031bc1e9f87ae1e74ad263382

manifests/input.yaml
ezhilsabareesh8: Update input and restart yaml files in this commit https://github.com/COSIMA/MOM6-CICE6/commit/058e5f7ee3ae4fd031bc1e9f87ae1e74ad263382

manifests/input.yaml
ezhilsabareesh8: Update input and restart yaml files in this commit https://github.com/COSIMA/MOM6-CICE6/commit/058e5f7ee3ae4fd031bc1e9f87ae1e74ad263382

manifests/input.yaml
ezhilsabareesh8: Update input and restart yaml files in this commit https://github.com/COSIMA/MOM6-CICE6/commit/058e5f7ee3ae4fd031bc1e9f87ae1e74ad263382

manifests/restart.yaml
```diff
@@ -1,3 +1,73 @@
 format: yamanifest
 version: 1.0
---- {}
```
anton-seaice, line 3: I think you committed this by accident? This file should be unchanged?

metadata.yaml
dougiesquire: Nope, we don't want to include this with a uuid specified since this could lead to multiple sets of model output with the same uuid. The om2 configs do include a template of the metadata.yaml for users to fill out, but let's just remove this for now

minghangli-uni:
> Shall we update ice_in to match the OM2 config now or in a later PR?
> Does @minghangli-uni have suggestions for performance changes or is that also a later PR?

I suggest so. For `0.25deg`, the processor shape is not recommended using `slenderX2`, but `square-ice` instead. Similar to distribution type, not `cartesian` but `roundrobin` instead. The above two changes omit land-only blocks and improve load balance. 

With below `ice.in` setting, 

```
&domain_nml
  block_size_x = 16
  block_size_y = 15
  distribution_type = "roundrobin"
  distribution_wght = "latitude"
  maskhalo_bound = .true.
  maskhalo_dyn = .true.
  maskhalo_remap = .true.
  max_blocks = 1000
  ns_boundary_type = "tripole"
  nx_global = 1440
  ny_global = 1080
  processor_shape = "square-ice"
```

The default `1deg` setting, 
```
&domain_nml
  block_size_x = 16
  block_size_y = 15
  distribution_type = "cartesian"
  distribution_wght = "latitude"
  maskhalo_bound = .true.
  maskhalo_dyn = .true.
  maskhalo_remap = .true.
  max_blocks = 1000
  ns_boundary_type = "tripole"
  nx_global = 1440
  ny_global = 1080
  processor_shape = "slenderX2"
```
With the updated `ice.in`, the speedup can be around 35%, calculated from the mean time of two consecutive runs.


> Shall we try and make a clean commit history to explain each change of somesort? (Or is this all one merge commit?)

A clean commit history is better for users to track and understand changes made.

anton-seaice:
Thanks @minghangli-uni . I think you can push that change straight to this branch, or if your prefer, make a PR into this branch for Ezhil to review first.

minghangli-uni:
Sure thing. I can make a PR into this branch soon.

manifests/restart.yaml
ezhilsabareesh8: Yes, that's a mistake. I restored it

config.yaml
ezhilsabareesh8: Reverted

config.yaml
ezhilsabareesh8: Reverted

MOM_input
```diff
@@ -1,7 +1,7 @@
 ! This input file provides the adjustable run-time parameters for version 6 of the Modular Ocean Model (MOM6).
 
 ! === module MOM ===
-DT = 1800.0
+DT = 1350.0
```
minghangli-uni, line 4: Should `DT_THERM` be an integer multiple of `DT`?

ezhilsabareesh8:
> For the 1deg we used initial conditions from 3 hour run of OM2, did we still want to generate those?

There's an ongoing discussion about the ice initial conditions as highlighted in [issue #50](https://github.com/COSIMA/access-om3/issues/50). We initially used the default option due to uncertainties in using OM2-generated initial conditions when switching to a C-grid, as discussed [here](https://github.com/COSIMA/access-om3/issues/50#issuecomment-2019428137).

The question remains: should we generate these initial conditions? or leave running the current configuration for a longer duration and examining the ice area output can provide insights into whether using OM2-generated initial conditions is needed or not.

MOM_input
ezhilsabareesh8: Ideally, `DT_THERM` should be an integer multiple of `DT` and less than the forcing or coupling time-step (DT = 1350.0). However, there are no warnings or errors with the current `DT_THERM` but we have to reduce it, any thoughts @aekiss ?

manifests/input.yaml
```diff
@@ -652,27 +647,27 @@ work/input/make_ryf/make_ryf.py:
...
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/mom/salt_sfc_restore.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/mom/salt_sfc_restore.nc
   hashes:
-    binhash: d1546a2b957a23e34f1154ad180df60d
-    md5: b2bd35c44017597ba99b85fb61ed4d72
+    binhash: abaea44efc3b0b31ab1a8eb9f46da744
+    md5: 434b4bdb323d17e9714d02519627fea2
 work/input/topog.nc:
-  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/1deg/share/topog.nc
+  fullpath: /g/data/ik11/inputs/access-om3/0.x.0/025deg/share/topog.nc
```
dougiesquire, line 670: There is also a `topog.nc` file in `/g/data/ik11/inputs/access-om3/0.x.0/025deg/mom`. This should be removed

dougiesquire (commented):
Thank @ezhilsabareesh8. A couple of initial comments. I'd still like to try and run the configuration, but I haven't had time today.

I think we maybe need to discuss whether we really want to be changing the coupling timestep here (and if so, what we want to do with MOM's thermodynamic timestep).

Also, from what I can tell the tool to generate the CICE grid isn't yet finialised. Should this wait until there is a single established tool?

dougiesquire:
@ezhilsabareesh8 I see you just force-pushed changes which has made Github think that some of my review comments are outdated. Please view my comments in this conversation, rather than in the "Files changed" tab, where a number of comments have disappeared.

anton-seaice:
> > For the 1deg we used initial conditions from 3 hour run of OM2, did we still want to generate those?
> 
> There's an ongoing discussion about the ice initial conditions as highlighted in [issue #50](https://github.com/COSIMA/access-om3/issues/50). We initially used the default option due to uncertainties in using OM2-generated initial conditions when switching to a C-grid, as discussed [here](https://github.com/COSIMA/access-om3/issues/50#issuecomment-2019428137).
> 
> The question remains: should we generate these initial conditions? or leave running the current configuration for a longer duration and examining the ice area output can provide insights into whether using OM2-generated initial conditions is needed or not.

I think we should generate the initial conditions, so that 
- the config is as close as possible to OM2 for verification of output, and 
- to reduce the number of potential stability issues we have during model spin-up (noting the same initial will probably used for CM3 initial set-up too).

minghangli-uni:
> But do we really want to change the coupling timestep? Is this done only because the CICE and coupling timesteps are [currently linked](https://github.com/COSIMA/access-om3/wiki/NUOPC-driver#cice6-time-steps)? Could we instead try to relax this requirement?

@dougiesquire In the current NUOPC setup, if the coupling timestep exceeds or equals 1800s, below two issues popped up,

1. `FATAL from PE 100: write energy: Ocean velocity has been truncated too many times`
2. `(abort ice) error = (diagnostic abort) ERROR: bad departure points`

From my perspective, the issues mainly stem from CICE, irrespective of the values assigned to `ndtd` during current tests, whether it's set to 1 or 2.

dougiesquire:
> I guess we could reduce the CICE timestep without reducing the coupling timestep? Is that what you are thinking?

Yeah, really just wondering whether that's even sensible. Probably not. And yeah, in OM2 the atmosphere-to-ice coupling is fixed at 3 hourly, but the ice-to-ocean coupling changes with the ocean-baroclinic/ice-thermodynamic timestep (which are the same). In MOM6 we have to think about what we want to do with `DT_THERM` - i.e. do we want this to be longer than the coupling timestep?

MOM_input
```diff
@@ -136,7 +155,7 @@ NJGLOBAL = 300
                                 ! The processor layout that was acutally used."
 
 ! === module MOM_verticalGrid ===
-NK = 50
+NK = 100
```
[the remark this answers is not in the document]
anton-seaice: I needed to make this suggested change for the MOM to start

dougiesquire:
> @dougiesquire In the current NUOPC setup, if the coupling timestep exceeds or equals 1800s, below two issues popped up

Okay, so maybe we don't really have a choice (at least for now). Changing the coupling timestep along with the ice and mom-baroclinic timesteps is certainly the easiest approach and aligns best with OM2. So the question that remains is what to do now with `DT_THERM`. In the long run, I presume we'll want this to be longer than the coupling timestep, so maybe now is the time to give `THERMO_SPANS_COUPLING` a whirl?

(P.S. the reason I'm so hung up on `THERMO_SPANS_COUPLING` is because I'll need to think about how this impacts WOMBAT coupling)

anton-seaice:
For reference, Kieran did test a patch to the CICE nuopc driver to run cice several times per coupling timesteps when he was messing around with the atmosphere coupling. It's looking like it won't be needed though.

anton-seaice:
Also - can someone make the 025deg_jra55do_ryf branch write protected?

minghangli-uni:
> Okay, so maybe we don't really have a choice (at least for now). Changing the coupling timestep along with the ice and mom-baroclinic timesteps is certainly the easiest approach and aligns best with OM2. So the question that remains is what to do now with DT_THERM. In the long run, I presume we'll want this to be longer than the coupling timestep, so maybe now is the time to give THERMO_SPANS_COUPLING a whirl?

These two [https://github.com/COSIMA/mom6-panan/issues/28](https://github.com/COSIMA/mom6-panan/issues/28) and  [https://github.com/jsimkins2/nwa25/tree/main/run](https://github.com/jsimkins2/nwa25/tree/main/run) could be useful for `dt_therm` selection. 

For a quick update, my current test case with `DT=1350` and `DT_THERM=5400` can run up to 2 years and is still running now.

dougiesquire:
> For a quick update, my current test case with DT=1350 and DT_THERM=5400 can run up to 2 years and is still running now.

Did you also set `THERMO_SPANS_COUPLING = True`?

minghangli-uni:
> Did you also set THERMO_SPANS_COUPLING = True? I'm guessing not because the NUOPC cap cannot handle THERMO_SPANS_COUPLING = True with DIABATIC_FIRST = True

No. `THERMO_SPANS_COUPLING` is in default False. and `DIABATIC_FIRST = False`

dougiesquire:
> No. THERMO_SPANS_COUPLING is in default False. and DIABATIC_FIRST = False

If you do not set `THERMO_SPANS_COUPLING = True` then your thermodynamic timestep is just getting reset internally to the baroclinic timestep. Also, have you changed the `DIABATIC_FIRST` setting because in this PR it is set to True?

minghangli-uni:
> have you changed the DIABATIC_FIRST setting because in this PR it is set to True?

It was discussed in [https://forum.access-hive.org.au/t/namelist-configuration-discussion-meeting/1917/8](https://forum.access-hive.org.au/t/namelist-configuration-discussion-meeting/1917/8) and agreed to better leave it in default first.

ezhilsabareesh8:
> > have you changed the DIABATIC_FIRST setting because in this PR it is set to True?
> 
> It was discussed in https://forum.access-hive.org.au/t/namelist-configuration-discussion-meeting/1917/8 and agreed to better leave it in default first.

Hi @minghangli-uni, did you change 'DIABATIC_FIRST' to true in the '1deg_jra55do_ryf' branch?. Since this PR is rebased from that.

MOM_input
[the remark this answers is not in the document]
ezhilsabareesh8: I downampled ocean_vgrid.nc to 51 interfaces and reverted this change in the latest commit

minghangli-uni:
For `1deg_jra55do_ryf`, it is True. But following the MOM6 parameter discussion, I havent run `1deg_jra55do_ryf`.

dougiesquire:
I think the `DIABATIC_FIRST` issue should be handled in a separate PR. Here, as in our other config branches, it is currently set to True.

anton-seaice:
Currently for the first day (if run for one day only):

Timer  11: ReadWrite       4.43 seconds
  Timer stats (node): min =        4.42 seconds
                      max =        4.43 seconds
                      mean=        4.42 seconds

After setting `pio_numiotasks = 5` for ICE:


Timer  11: ReadWrite       3.30 seconds
  Timer stats (node): min =        3.30 seconds
                      max =        3.30 seconds
                      mean=        3.30 seconds

This number can get tuned after the number the main run set-up is tuned for performance.

MOM_input
```diff
@@ -400,7 +419,7 @@ OCEAN_SURFACE_STAGGER = A
                                 ! returned to the coupler.  Valid values include
                                 ! 'A', 'B', or 'C'."
 
-RESTART_CONTROL = 3
+RESTART_CONTROL = 1
```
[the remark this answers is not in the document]
ezhilsabareesh8: It was agreed to change `RESTART_CONTROL =1` in the MOM parameter discussion meeting,  [refer here](https://forum.access-hive.org.au/t/namelist-configuration-discussion-meeting/1917/9#:~:text=RESTART_CONTROL%20%2D%20try%20using%20default%20(1))

nuopc.runconfig
```diff
@@ -284,23 +284,23 @@ CLOCK_attributes::
...
      stop_n = 1
-     stop_option = nmonths
+     stop_option = nyears
      stop_tod = 0
      stop_ymd = -999
      tprof_n = -999
      tprof_option = never
      tprof_ymd = -999
-     wav_cpl_dt = 3600
+     wav_cpl_dt = 1800
```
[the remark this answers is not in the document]
ezhilsabareesh8: `wav_cpl_dt` is not used in this configuration, maybe we can remove it?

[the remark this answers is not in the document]
anton-seaice: If removing it works, then please make a new issue so we do it in all the relevant configs :)

dougiesquire:
> This number can get tuned after the number the main run set-up is tuned for performance.

No comprendo

nuopc.runconfig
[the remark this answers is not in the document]
ezhilsabareesh8: `wav_cpl_dt` is changed to the default value

anton-seaice:
> > This number can get tuned after the number the main run set-up is tuned for performance.
> 
> No comprendo

Ah I just mean that figuring out what the best value of `pio_numiotasks` is depends on the final number of nodes chosen (which comes out of doing performance tests/runs), and possibly the chunksizes set for netcdf. Therefore its not worth worrying about too much just yet until we refine those things.

metadata.yaml
```diff
@@ -0,0 +1,2 @@
+experiment_uuid: 071afe24-6659-4385-baa0-28e6eb1f588e
+parent_experiment: 76c010c8-6f80-4600-bbd2-3135bed38cd5
```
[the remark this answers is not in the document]
ezhilsabareesh8: Removed

MOM_input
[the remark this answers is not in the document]
dougiesquire: Do you know what this does? I haven't looked in detail but it looks like it just turns off timestamping the restart files, which I think would do nothing for us since the NUOPC cap sets the restart filename.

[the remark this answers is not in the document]
ezhilsabareesh8: Yes it is for time stamping and in our configs NUOPC does that, so we agreed to restore it to a default value which is 1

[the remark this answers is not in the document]
anton-seaice: Can we delete it ? :) Or put at worst, put a comment in MOM_input that it's not used?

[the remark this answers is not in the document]
dougiesquire: Yeah, delete it

minghangli-uni:
> This is the updated &domain_nml in ice.in, which is consistent with OM2.
```
&domain_nml
  block_size_x = 16
  block_size_y = 15
  distribution_type = "roundrobin"
  distribution_wght = "latitude"
  maskhalo_bound = .true.
  maskhalo_dyn = .true.
  maskhalo_remap = .true.
  max_blocks = 1000
  ns_boundary_type = "tripole"
  nx_global = 1440
  ny_global = 1080
  processor_shape = "square-ice"
```

https://github.com/COSIMA/access-om3/issues/130#issuecomment-2038822573: A large value of `max_blocks` will not cause the code to fail, but will allocate more memory than is necessary, and hence affect scalability. Reduce it to 0 (not sure now) or an appropriate value will improve scalability.

anton-seaice:
> [COSIMA/access-om3#130 (comment)](https://github.com/COSIMA/access-om3/issues/130#issuecomment-2038822573): A large value of `max_blocks` will not cause the code to fail, but will allocate more memory than is necessary, and hence affect scalability. Reduce it to 0 (not sure now) or an appropriate value will improve scalability.

Have you tested it?

minghangli-uni:
I haven't. But @micaeljtoliveira did. 

> Setting max_blocks to a more reasonable value for each core count leads to improved scalability and the percentage of time spent in the surface forcing routines does not increase as much.

ezhilsabareesh8:
> Reduce it to 0 (not sure now) or an appropriate value will improve scalability.

The value should be -1 to compute `max_blocks` on the fly, check [this](https://cice-consortium-cice.readthedocs.io/en/cice6.0.1/user_guide/ug_implementation.html#:~:text=The%20code%20will%20also%20print%20a%20warning%20if%20the%20maximum%20number%20of%20blocks%20is%20too%20large.%20Although%20this%20is%20not%20fatal%2C%20it%20does%20use%20extra%20memory.%20If%20max%5C_blocks%20is%20set%20to%20%2D1%2C%20the%20code%20will%20compute%20a%20max%5C_blocks%20on%20the%20fly)

anton-seaice:
Ah thankyou both! Ill put that in a seperate change.

dougiesquire:
It looks setting `max_blocks` to anything < 1, will [internally do](https://github.com/ESCOMP/CICE/blob/64177e3e5658711f5f19030ddfcc6927f1dbda5d/cicecore/cicedyn/infrastructure/ice_domain.F90#L220-L226):

```fortran
max_blocks=( ((nx_global-1)/block_size_x + 1) *         &
             ((ny_global-1)/block_size_y + 1) - 1) / nprocs + 1
max_blocks=max(1,max_blocks)
```

micaeljtoliveira:
> The value should be -1 to compute max_blocks on the fly, check [this](https://cice-consortium-cice.readthedocs.io/en/cice6.0.1/user_guide/ug_implementation.html#:~:text=The%20code%20will%20also%20print%20a%20warning%20if%20the%20maximum%20number%20of%20blocks%20is%20too%20large.%20Although%20this%20is%20not%20fatal%2C%20it%20does%20use%20extra%20memory.%20If%20max%5C_blocks%20is%20set%20to%20%2D1%2C%20the%20code%20will%20compute%20a%20max%5C_blocks%20on%20the%20fly)

I tried it with the latest CICE and the previous version we were using and it does not seem to work correctly. I always get this error message:
```
 (abort_ice) error = 
 (create_distrb_cart)ERROR: max_blocks too small (need at least 10)
 ERROR: 
 (abort_ice)(create_distrb_cart)ERROR: max_blocks too small (need at least 10)
```

anton-seaice:
Aah .. we don't set nprocs in `ice_in` (because nuopc), so that max_blocks calculate will always be wrong. I guess Ill raise an issue!

micaeljtoliveira:
Am I allowed to set `nprocs` in `ice_in` (provided it's consistent with the corresponding value in nuopc)?

anton-seaice:
Yes. I just tried that and it works. Delete the max_blocks line and add the nprocs = 48 (or whatever the magic number is)

micaeljtoliveira:
> Yes. I just tried that and it works. Delete the max_blocks line and add the nprocs = 48 (or whatever the magic number is)

I've also tried for different values of nproc and it works in a few cases, but fails in general. I think the issue is that the algorithm @dougiesquire copied above is not valid in general for the slenderX2 distribution.

dougiesquire:
> Multiple restart files were being generated by MOM6 due to netCDF file limits. This issue has been addressed and fixed in this https://github.com/payu-org/payu/pull/432.

Note, this PR has now been merged, so this will be fixed with the next Payu release

nuopc.runconfig
```diff
@@ -284,17 +284,17 @@ CLOCK_attributes::
      glc_avg_period = yearly
      glc_cpl_dt = 86400
      history_ymd = -999
-     ice_cpl_dt = 3600
+     ice_cpl_dt = 1350
      lnd_cpl_dt = 3600
-     ocn_cpl_dt = 3600
+     ocn_cpl_dt = 1350
```
minghangli-uni, line 289: A quick note on these. As was mentioned in [https://github.com/COSIMA/access-om3/wiki/NUOPC-driver#time-steps](https://github.com/COSIMA/access-om3/wiki/NUOPC-driver#time-steps), for our current NUOPC config, `{model_name}_cpl_dt` are unused. 

> So I think in our case, {model_name}_cpl_dt are unused and the driver time-step equals the coupling time-step set in nuopc.runseq. Certainly, changing these values seems to have no effect.

dougiesquire: Yeah. I wonder if we should change these to something obviously ridiculous (in another PR) to make it clear to our future selves that they aren't used. E.g. I wrote that documentation in the wiki, but I forgot that they aren't used when I was reviewing this PR... 😅

dougiesquire: I've open this issue so we can come back to this: https://github.com/COSIMA/access-om3/issues/132

MOM_input
ezhilsabareesh8: This issue will be handled in a separate PR as discussed [here](https://github.com/COSIMA/MOM6-CICE6/pull/48#issuecomment-2038431642)

dougiesquire: I think we need to decide what to do about `DT_THERM` here. The current settings are confusing:
`DT_THERM > DT` but `THERMO_SPANS_COUPLING = False (by default)`.

With these settings, I think **the value of `DT_THERM`  in the `MOM_input` will not be used** - instead `DT_THERM` will be reset internally to `DT`. If we do want to have `DT_THERM > DT` we need to set `THERMO_SPANS_COUPLING = True` which will require we also change to `DIABATIC_FIRST = False` due to limitations in the MOM6 NUOPC cap.

Thoughts @aekiss

dougiesquire: Options are:

1. Change to `DT_THERM = 1350.0` (the coupling timestep) and leave other setting as they are.
   
2. Leave `DT_THERM` as it is (`3600.0`) and change to `THERMO_SPANS_COUPLING = True` and `DIABATIC_FIRST = False`

MOM_input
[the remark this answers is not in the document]
ezhilsabareesh8: `ocean_vgrid.nc` is downsampled and changes to `NK` has been reverted

MOM_input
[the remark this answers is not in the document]
ezhilsabareesh8: Deleted in this commit acbb068

MOM_input
aekiss: It's certainly confusing. Do we set `single_step_call` anywhere? It defaults to true, in which case `DT`, `DT_THERM` and `THERMO_SPANS_COUPLING` are ignored, [according to this](https://github.com/NCAR/MOM6/blob/13fbeb7/config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90#L168-L178).

Additional relevant documentation in MOM6 cap is [here](https://github.com/NCAR/MOM6/blob/13fbeb7/config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90#L293-L318).

aekiss: As discussed in today's meeting, let's do [option 1](https://github.com/COSIMA/MOM6-CICE6/pull/48#discussion_r1555126419) for now.

We can then explore option 2 with different `DT_THERM` in subsequent tests and PRs. (Note that when `THERMO_SPANS_COUPLING = True`, `DT_THERM` [will be rounded down to an integer number of coupling timesteps](https://github.com/NCAR/MOM6/blob/13fbeb7/config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90#L309-L314).)

aekiss: Given how confusingly these parameters interact, it will be good to have `MOM_parameter_doc.*` added to the repos to provide a sanity check on what is actually being used for the test when we do option 2 - see https://github.com/COSIMA/access-om3/issues/117. We could do this manually until CI is in place to automate it.

dougiesquire:
We discussed this PR in a meeting today. We agreed to:

- set `DT_THERM = 1350.0` for now. This will give us a working configuration to begin testing changing `DT_THERM` to something longer than the coupling timestep.
- leave the ice initial conditions as `default`
- In a subsequent PR, set the `<model>_cpl_dt` parameters in the `nuopc.runconfig` to something obvious ridiculous and add a comment (if possible) that these are not used. We'll need to do it for all configs. I've open an issue for this here: https://github.com/COSIMA/access-om3/issues/132

@ezhilsabareesh8 can you make these changes and ping us for another review.

MOM_input
aekiss: If `DT_THERM` exceeds the coupling timestep, but the coupling timestep matches the dynamic timestep in CICE and baroclinic `DT` in MOM6, would it make sense to set the thermo timestep in CICE to match that in MOM?

This would involve setting `ndtd` in CICE to `DT_THERM/DT` and setting the CICE thermo timestep to `ndtd*DT`.

dougiesquire: Open question: can CICE run with a thermo timestep greater than the coupling timestep?

aekiss:
Also see comments [above](https://github.com/COSIMA/MOM6-CICE6/pull/48#discussion_r1558583441) from today.

ezhilsabareesh8:
> @ezhilsabareesh8 can you make these changes and ping us for another review.

I have made these changes in commit [ff2ef98](https://github.com/COSIMA/MOM6-CICE6/pull/48/commits/ff2ef98d31db4c2fdb1808e08f7642e5a6b84ead) and [b44852b](https://github.com/COSIMA/MOM6-CICE6/pull/48/commits/b44852b554c66271eafb6df9b594ad21663fe29d)

dougiesquire:
> I have made these changes in commit [ff2ef98](https://github.com/COSIMA/MOM6-CICE6/pull/48/commits/ff2ef98d31db4c2fdb1808e08f7642e5a6b84ead) and [b44852b](https://github.com/COSIMA/MOM6-CICE6/pull/48/commits/b44852b554c66271eafb6df9b594ad21663fe29d)

I think we should leave [b44852b](https://github.com/COSIMA/MOM6-CICE6/pull/48/commits/b44852b554c66271eafb6df9b594ad21663fe29d) for a subsequent PR since it really should be done across all configs

aekiss:
A subsequent PR should also upgrade exes https://github.com/COSIMA/access-om3/issues/118

aekiss:
As a cross-check, here are the `MOM_input` differences between this PR and the other latest versions of MOM6-CICE6: 
https://github.com/COSIMA/access-om3-doc/blob/main/tables/MOM_input/MOM6-CICE6_nml_diff.md
and the `ice_in` differences, also comparing to ACCESS_OM2-025: 
https://github.com/COSIMA/access-om3-doc/blob/main/tables/ice_in/cice_1_025_nml_diff.md

ezhilsabareesh8:
> I think we should leave [b44852b](https://github.com/COSIMA/MOM6-CICE6/pull/48/commits/b44852b554c66271eafb6df9b594ad21663fe29d) for a subsequent PR since it really should be done across all configs

Thanks @dougiesquire , I have reverted it to original cpl dt values in this commit 043f97b

aekiss (approved):
Looks good to me - thanks @ezhilsabareesh8 and all

manifests/input.yaml
dougiesquire: I've removed this

dougiesquire:
@ezhilsabareesh8 I'm just doing a test run with this PR and if that's successful I'll approve. Are you happy for me to rebase this and deal with the conflicts?

aekiss:
Do we also want to merge this or do that as a separate PR? 
https://github.com/COSIMA/MOM6-CICE6/tree/025deg_jra55do_ryf_iss135
https://github.com/COSIMA/access-om3/issues/135

ezhilsabareesh8:
> I'm just doing a test run with this PR and if that's successful I'll approve. Are you happy for me to rebase this and deal with the conflicts?

Thanks @dougiesquire, yes that would be great.

dougiesquire:
> Do we also want to merge this or do that as a separate PR?
https://github.com/COSIMA/MOM6-CICE6/tree/025deg_jra55do_ryf_iss135
https://github.com/COSIMA/access-om3/issues/135

I think the changes in https://github.com/COSIMA/MOM6-CICE6/tree/025deg_jra55do_ryf_iss135 that aren't already included in this PR should be handled in separate PR(s). @minghang, can you have another look at the changes here to check that you're happy with them before we merge. The CICE `block_size_x`, `block_size_y` and `max_blocks`, in particular, are different than what you propose in https://github.com/COSIMA/access-om3/issues/135

minghangli-uni:
Hi @dougiesquire My account is @minghangli-uni instead of @minghang, which you have been referring to. 

Regarding the parameters `block_size_x=30` and `block_size_y=27`, these values were directly grabbed from the [OM2 technical report](https://github.com/COSIMA/ACCESS-OM2-1-025-010deg-report). And `max_blocks=8` is calculated from [this code](https://github.com/ESCOMP/CICE/blob/64177e3e5658711f5f19030ddfcc6927f1dbda5d/cicecore/cicedyn/infrastructure/ice_domain.F90#L220-L226). As I used 288 cores, the corresponding value for `max_blocks` was 8. However, for the current config, which employs 240 cores, `max_blocks` should be set to 10.

Given that we have reliable references to validate these parameter values, I prefer to maintain `block_size_x=30`, `block_size_y=27`, and `max_blocks=10`.

I am happy for the rest of the change.

dougiesquire:
> My account is @minghangli-uni instead of @minghang, which you have been referring to.

Whoops, sorry!

minghangli-uni:
No problem at all. I'm concerned about you quoting the wrong person, and I might not receive your message in time.

anton-seaice:
> Regarding the parameters `block_size_x=30` and `block_size_y=27`, these values were directly grabbed from the [OM2 technical report](https://github.com/COSIMA/ACCESS-OM2-1-025-010deg-report). And `max_blocks=8` is calculated from [this code](https://github.com/ESCOMP/CICE/blob/64177e3e5658711f5f19030ddfcc6927f1dbda5d/cicecore/cicedyn/infrastructure/ice_domain.F90#L220-L226). As I used 288 cores, the corresponding value for `max_blocks` was 8. However, for the current config, which employs 240 cores, `max_blocks` should be set to 10.

The larger block sizes seem like a more sensible default :)

Feel free to mess with the numbers as much as you would like during the profiling work. The cice doco states: 

`
 In practice, blocks should probably not have fewer than about 8 to 10 grid cells in each direction, and more square blocks tend to optimize the volume-to-surface ratio important for communication cost. Often 3 to 8 blocks per processor provide the decompositions flexiblity to create reasonable load balance configurations.
`

So moving to bigger blocks rather than increasing max_blocks might make sense.

Also see https://cice-consortium-cice.readthedocs.io/en/main/user_guide/ug_implementation.html#fig-distribscorecard . I don't know why we use 'sectrobin' either.

Anyway - this is for future work, not this PR :)

minghangli-uni:
We use `roundrobin`

https://github.com/COSIMA/MOM6-CICE6/blob/185b44f46f10aa4700fd3b094dfa7fec9484f1e9/ice_in#L70

dougiesquire:
> @ezhilsabareesh8 I'm just doing a test run with this PR and if that's successful I'll approve.

So this ran for me, but it took 19 hours to complete 1 year-long run. Is that expected?

minghangli-uni:
Yes, with the current configuration, using `DT=DT_THERM=1350s`, a year-long run took approximately 19 hours to complete.

dougiesquire (approved):
Thanks @ezhilsabareesh8 et al. LGTM and ran as is.

anton-seaice:
Can this get merged ?

dougiesquire:
@ezhilsabareesh8 we're happy for you to merge

ezhilsabareesh8:
> @ezhilsabareesh8 we're happy for you to merge

Thanks @dougiesquire , I will merge then

PR #53, opened 2024-04-05
1deg_jra55do_ryf: update OM3 executable to latest version

micaeljtoliveira:
Changes needed to use latest OM3 executable.

Contributes to https://github.com/COSIMA/access-om3/issues/118

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -18,7 +18,7 @@ jobname: 1deg_jra55do_ryf
 
 model: access-om3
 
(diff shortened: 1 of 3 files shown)


dougiesquire (approved):
LGMT and ran successfully

micaeljtoliveira:
Marking this as a draft, as I'll have to update the executable again, as I need to reinstall OM3 after https://github.com/COSIMA/access-om3/pull/131

micaeljtoliveira:
Merging this as is. I'm going to change the way how we specify the executable path so that we don't have to update all configurations whenever we update the development executable and I prefer to do it in a separate PR.

PR #54, opened 2024-04-05
CICE config for CICE6.5

anton-seaice:
Changes described in https://github.com/COSIMA/access-om3/issues/113 for move to CICE 6.5. Uncompressed daily history output is 15MB, compressed is 2MB.

Add new cice grid from https://github.com/COSIMA/esmgrids/pull/6 and (https://github.com/COSIMA/access-om2/issues/279 )

Tigthen grid angle check limits (per https://github.com/COSIMA/access-om3/issues/144)

Add freeboard as a default history output to make validation easier.

Turn off metadata for development work.

Change should be described in each commit, I will not squash the commits.

### DIFF
--- a/ice_in
+++ b/ice_in
@@ -127,6 +127,10 @@
   !-----------------------------------
   ! These fields are on by default (in ice_history_shared.F90) but lets turn them off
   !-----------------------------------
+  f_tlon         = .false. , f_tlat         = .false.
+  f_ulon         = .false. ,  f_ulat         = .false.
+  f_nlon         = .false. ,  f_nlat         = .false.
+  f_elon         = .false. ,  f_elat         = .false.
   f_albpnd = 'x'
   f_atmdir = 'x' , f_atmspd = 'x'
   f_coszen = 'x'
--- a/ice_in
+++ b/ice_in
@@ -7,6 +7,7 @@
   histfreq = "d", "m", "x", "x", "x"
   hist_time_axis = "middle"
   history_deflate = 1
+  history_chunksize = 180, 150
   history_precision = 8
   restart_deflate = 1
   ice_ic = "./input/iced.1900-01-01-10800.nc"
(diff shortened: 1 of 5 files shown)


anton-seaice:
@aekiss @dougiesquire Can you review please? :)

ice_in
```diff
@@ -124,12 +125,26 @@
...
+  f_elon         = .false. , f_elat         = .false.
+  f_tmask        = .false. , f_umask        = .false.
+  f_nmask        = .false. , f_emask        = .false.
+  f_tarea        = .false. , f_uarea        = .false.
+  f_narea        = .false. , f_earea        = .false.
+  f_dxt          = .false. , f_dyt          = .false.
+  f_dxu          = .false. , f_dyu          = .false.
+  f_dxe          = .false. , f_dye          = .false.
+  f_dxn          = .false. , f_dyn          = .false.
+  f_HTN          = .false. , f_HTE          = .false.
```
aekiss, line 147: with compression these will be very small, but I take your point about them not being very useful and slowing it down so I'm happy for them to be false

aekiss (approved):
Thanks @anton-seaice - LGTM, sorry for the very slow turnaround

ice_in
anton-seaice: In xarray, when you do a concatenation operation, it does a check that the coordinates are the same in all the datasets. This is very slow, hence the reason for removing them.

anton-seaice:
> Thanks @anton-seaice - LGTM, sorry for the very slow turnaround

Not slow at all :) I only requested it yesterday :)

dougiesquire (approved):
LGTM and runs as is, thanks @anton-seaice. Most of these changes are also needed in our other configs. I presume that's the plan?

anton-seaice:
Thanks dougie. Yes I will go through and cherry-pick these into the other configs and check if there are differences needed.

PR #57, opened 2024-04-18
1deg_jra55do_iaf: Update OM3 executable to latest version

micaeljtoliveira:
This implements a new naming scheme for the exec path.

Contributes to https://github.com/COSIMA/access-om3/issues/118

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -17,8 +17,7 @@ walltime: 01:00:00
 jobname: 1deg_jra55do_iaf
 
 model: access-om3
-
-exe: /g/data/ik11/spack/0.21.2/opt/linux-rocky8-cascadelake/intel-2021.10.0/access-om3-main-6pue372/bin/access-om3-MOM6-CICE6
+exe: /g/data/ik11/spack/0.21.2/opt/linux-rocky8-cascadelake/intel-2021.10.0/access-om3-d6813d6b9e1df560ac3f6ba6a605daab9cfd9569_main-5pjh7z2/bin/access-om3-MOM6-CICE6
 input: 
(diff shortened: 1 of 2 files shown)


dougiesquire (approved):
Great - nice to have the git tag so easy to find

dougiesquire:
Actually, can you hold off on merging. Should probably do a test run

dougiesquire:
Ran successfully 👍

micaeljtoliveira:
> Thanks @micaeljtoliveira, looks good to me. But it could be neater and easier to visually compare with the github commit history if you used a short hash (7 chars).
> 
> Would it be useful to also include the commit date?
> 
> I guess 5pjh7z2 is the spack hash, but what purpose does "main" serve? Could drop if it's redundant.

I'm afraid we don't have much control over all this. The information that can be used by Spack in install paths is very limited (compiler, architecture, package name, version, spack hash). So no date here.

This part of the path `access-om3-d6813d6b9e1df560ac3f6ba6a605daab9cfd9569_main-5pjh7z2` corresponds to `{package-name}-{package-version}-{spack-hash-7}`. 

`main` is part of the version and is the package version that corresponds to this commit. You can read all about it [here](https://spack.readthedocs.io/en/latest/basic_usage.html#git-versions), but to make it short, this is necessary and the default used by spack is even worse (here it would be something like `0.2.0.git.20`, which means "20 commits after tag 0.2.0").

I think one could shorten the git commit hash, but then people might easily confused it with the spack hash, so I'm not sure that's a good idea.

micaeljtoliveira:
@aekiss I think for now this is the best we can do (see explanation above). I'll proceed with merging so we can move forward. If needed, the length of the git hash can be shortened next time we install a new version.

PR #74, opened 2024-06-19
Manually add auto-generated MOM_parameter_doc.* for comprehensive documentation

minghangli-uni:
https://github.com/COSIMA/access-om3/issues/174

### DIFF
--- a/docs/MOM_parameter_doc.all
+++ b/docs/MOM_parameter_doc.all
@@ -1,0 +1,2336 @@
+! This file was written by the model and records all non-layout or debugging parameters used at run-time.
+
+! === module MOM ===
+SPLIT = True                    !   [Boolean] default = True
+                                ! Use the split time stepping if true.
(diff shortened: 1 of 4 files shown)


dougiesquire (changes requested):
Thanks @minghangli-uni. Ideally [these will be automatically generated](https://github.com/COSIMA/access-om3/issues/117), so we don't have to stay on top of generating them, but I guess this will have to do for now. But I think we should move these into a `docs` subdirectory so that it's clear they are not part of the configuration.

dougiesquire (approved):
Thanks @minghangli-uni. LGTM (note I haven't actually checked that the parameter docs are correct for the configuration)

minghangli-uni:
Thanks @dougiesquire. These log files are not used by the MOM6, hence even if they lack some parameters from `MOM_input` (this [bug](https://github.com/COSIMA/access-om3/issues/121)), they will not affect our results. I will monitor the changes throughout the process, so rest assured.

PR #76, opened 2024-06-21
[0.25deg] Change MOM6 tracer timestep

minghangli-uni:
This PR changes the MOM6 tracer timestep within the MOM module so that the curent DT_THERM (8100s) is 6 times longer than DT (1350s). The coupling timestep remains the same as DT at this stage. 

Evaluations and scientific tests are required before merging.

See https://github.com/COSIMA/access-om3/issues/138#issuecomment-2179999777

### DIFF
--- a/MOM_input
+++ b/MOM_input
@@ -1,9 +1,6 @@
 ! This file was written by the model and records the non-default parameters used at run-time.
 
 ! === module MOM ===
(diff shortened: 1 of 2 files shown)


minghangli-uni:
This will be closed. See https://github.com/ACCESS-NRI/access-om3-configs/pull/223

PR #77, opened 2024-06-21
[0.25deg] Disable TIDES and associated parameters

minghangli-uni:
Disable `TIDES` and associated parameters. One thing to note: we might need to add the `INT_TIDE_DISSIPATION` parameter to parameterise the turbulent mixing generated by breaking internal waves for local dissipation. To do this, `tideamp.nc`, which contains the averaged tidal flow amplitudes must be included in the input files.

See https://github.com/COSIMA/access-om3/issues/176

### DIFF
--- a/docs/MOM_parameter_doc.all
+++ b/docs/MOM_parameter_doc.all
@@ -1112,7 +1112,7 @@ PORBAR_ETA_INTERP = "MAX"       ! default = "MAX"
                                 !    HARMONIC - harmonic mean of the adjacent cells
 
 ! === module MOM_dynamics_split_RK2 ===
-TIDES = True                    !   [Boolean] default = False
+TIDES = False                   !   [Boolean] default = False
                                 ! If true, apply tidal momentum forcing.
 BE = 0.6                        !   [nondim] default = 0.6
                                 ! If SPLIT is true, BE determines the relative weighting of a  2nd-order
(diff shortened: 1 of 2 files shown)


docs/MOM_parameter_doc.all
```diff
@@ -1255,7 +1255,7 @@ TIDAL_SAL_FROM_FILE = False     !   [Boolean] default = False
 USE_PREVIOUS_TIDES = False      !   [Boolean] default = False
                                 ! If true, use the SAL from the previous iteration of the tides to facilitate
                                 ! convergent iteration. This is only used if TIDES is true.
-TIDE_USE_SAL_SCALAR = True      !   [Boolean] default = True
+TIDE_USE_SAL_SCALAR = False     !   [Boolean] default = True
```
aekiss, line 1258:
```suggestion
TIDE_USE_SAL_SCALAR = True     !   [Boolean] default = True
```
Set to default for consistency with `MOM_input` (value is ignored since TIDES is false)

aekiss (approved):
LGTM, thanks @minghangli-uni - just one suggested change

docs/MOM_parameter_doc.all
minghangli-uni: Thanks for the suggestion. It has been fixed.

PR #79, opened 2024-06-21
[0.25deg] Add both ZSTAR and rho2 in DIAG_COORDS

minghangli-uni:
See https://github.com/COSIMA/access-om3/issues/178


minghangli-uni:
Current goal is not including rho2, hence this PR will be closed now.

PR #80, opened 2024-06-26
Modify submesoscale parameterisation

minghangli-uni:
See https://github.com/COSIMA/access-om3/issues/181 for more discussions.


minghangli-uni:
This is closed because of outdated parameters.

PR #84, opened 2024-07-09
Update on mesoscale parameterisation. 

minghangli-uni:
Only GM (MEKE is off) is used for thickness diffusivity. It is formulated as a boundary value problem following Ferrari et al. (2010), similar to OM2, though there are some differences in parameters. The diffusivity remains depth-independent but is flow-dependent, using Visbeck et al. (1997), akin to OM2. The primary differences in flow-dependent diffusivity lie in the Eady growth rate and the scaling factor. 

For the neutral tracer diffusion, the tracer diffusivity follows the same approach as OM2, with a maximum cap of 200 m2/s. A resolution function is also applied.

See https://github.com/COSIMA/access-om3/issues/179


minghangli-uni:
This will be closed because of some of parameters in this PR are outdated.

PR #85, opened 2024-07-09
Turn on AUTO_MASKTABLE 

minghangli-uni:
to automatically generate a mask table to exclude land masks.

See https://github.com/COSIMA/access-om3/issues/184

### DIFF
--- a/docs/MOM_parameter_doc.layout
+++ b/docs/MOM_parameter_doc.layout
@@ -23,9 +23,9 @@ THIN_HALO_UPDATES = True        !   [Boolean] default = True
                                 ! statically determined at compile time.  Otherwise the sizes are not determined
                                 ! until run time. The STATIC option is substantially faster, but does not allow
                                 ! the PE count to be changed at run time.  This can only be set at compile time.
-AUTO_MASKTABLE = False          !   [Boolean] default = False
+AUTO_MASKTABLE = True           !   [Boolean] default = False
                                 ! Turn on automatic mask table generation to eliminate land blocks.
-MASKTABLE = "MOM_mask_table"    ! default = "MOM_mask_table"
+MASKTABLE = "MOM_auto_mask_table"    ! default = "MOM_mask_table"
                                 ! A text file to specify n_mask, layout and mask_list. This feature masks out
                                 ! processors that contain only land points. The first line of mask_table is the
                                 ! number of regions to be masked out. The second line is the layout of the model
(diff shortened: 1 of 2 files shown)


docs/MOM_parameter_doc.layout
```diff
@@ -46,6 +46,10 @@ NJPROC = 30                     !
                                 ! in MOM_memory.h at compile time.
 LAYOUT = 36, 30                 !
                                 ! The processor layout that was actually used.
+AUTO_IO_LAYOUT_FAC = 0          ! default = 0
+                                ! When AUTO_MASKTABLE is enabled, io layout is calculated by performing integer
+                                ! division of the runtime-determined domain layout with this factor. If the
+                                ! factor is set to 0 (default), the io layout is set to 1,1.
 IO_LAYOUT = 1, 1                ! default = 1
```
anton-seaice, line 53: It looks like `IO_Layout` is not used? Can we delete it? It is determined by `AUTO_IO_LAYOUT_FAC`? 

Per https://gmd.copernicus.org/articles/13/1885/2020/ it looks like the default is the right value, but is performance impacted if you increase AUTO_IO_LAYOUT_FAC ?

anton-seaice:
Slightly off-topic ... shall we mark these lines as unused while we are at it?

> OCN_modelio::
>      ...
>      pio_async_interface = .false.
>      pio_netcdf_format = 64bit_offset
>      pio_numiotasks = -99
>      pio_rearranger = 2
>      pio_root = 1
>      pio_stride = 48
>      pio_typename = netcdf
> ::

dougiesquire:
> Slightly off-topic ... shall we mark these lines as unused while we are at it?
> 
> > OCN_modelio::
> > ...
> > pio_async_interface = .false.
> > pio_netcdf_format = 64bit_offset
> > pio_numiotasks = -99
> > pio_rearranger = 2
> > pio_root = 1
> > pio_stride = 48
> > pio_typename = netcdf
> > ::

Can we open a separate issue for this? If we do this, we should do it for all configs at once.

docs/MOM_parameter_doc.layout
minghangli-uni: `MOM_parameter_doc.layout` is an auto-generated file, so there is no need to manually delete any parameters in it. The I/O layout has not been tuned yet, but it will definitely change when varying `AUTO_IO_LAYOUT_FAC`.

anton-seaice:
Ill test this now

anton-seaice (approved):
This starts and runs.

It produces a file "MOM_auto_mask_table" which has some numbers in it, not sure what they mean :)

minghangli-uni:
Thank you @anton-seaice 

In `MOM_auto_mask_table`, the numbers on both columns indicate the 2D processors in the x and y directions that are to be masked out.

PR #86, opened 2024-07-12
Update on lateral friction

minghangli-uni:
Lateral friction is implemented using a biharmonic operator, with a Smagorinsky scaling parameter of 0.05066, which is equivalent to the isotropic scaling parameter of [2.0 in OM2](https://github.com/COSIMA/025deg_jra55_ryf/blob/0b4f66537484bbf37d6abce539f6857cbf745897/ocean/input.nml#L121). Additionally, a static biharmonic viscosity is applied as referenced in [GFDL OM4 & OM5](https://github.com/COSIMA/access-om3-doc/blob/19bc95c1aaad87cee3f653d5197efeef4cb42191/configs/GFDL-OM5/b03_update_MOM_parameter_doc.all#L1348-L1351). Unlike OM2, MOM6 implements a slip BC for lateral velocity.

See https://github.com/COSIMA/access-om3/issues/186

### DIFF
--- a/MOM_input
+++ b/MOM_input
@@ -372,15 +372,16 @@ MASS_WEIGHT_IN_PRESSURE_GRADIENT = True !   [Boolean] default = False
                                 ! bathymetry in FV pressure gradient calculations.
 
 ! === module MOM_hor_visc ===
+SMAGORINSKY_AH = True           !   [Boolean] default = False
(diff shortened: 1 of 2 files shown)


minghangli-uni:
This PR will be closed because of outdated parameters.

PR #102, opened 2024-08-21
1deg jra55do iaf.cherrypicks

anton-seaice:
companion PR to #91


dougiesquire (approved):
Thanks @anton-seaice! LGTM and verified runs successfully as is

PR #103, opened 2024-08-21
025deg jra55do ryf.cherrypicks

anton-seaice:
Companion PR to #91 

This fails as there is no `input/iced.1900-01-01-10800.nc` in this configuration, as we use `ice_ic=default`. The iced file is expected by the setup user script.

We could just not use the setup user script, or change the user script, or make the initial condition file. Thoughts @dougiesquire ?


dougiesquire:
I think ditch the setup user script here. That's what we'll do for the other configs anyway once the symlink workaround is no longer required.

If you think it's important to make an initial condition file, that should be done in a subsequent PR

anton-seaice:
I ran two adhoc tests:
- with run length of 1 day, I did three 1 day runs to confirm that this reads restart files. From ice.log:
```
 Read ./rpointer.ice
 Using restart dump=./access-om3.cice.r.1900-01-02-00000.nc
 (ice_pio_init) opening file for reading ./access-om3.cice.r.1900-01-02-00000.nc
Restart read at istep=      64    1900-01-02-00000
```

- with run length of 1 month, it concats the daily ice history output

config.yaml
```diff
@@ -38,5 +38,13 @@ metadata:
     enable: false
 
 userscripts:
-    setup: ./setup_cice_restarts.sh
-    archive: ./archive_cice_restarts.sh
+    setup: if [ -f work/access-om3.cice.r.* ] ; then RESTART=$(echo work/access-om3.cice.r.*) ; ln -f $(readlink $RESTART) $RESTART ; fi
```
dougiesquire, line 41: Sorry, I forgot that the setup script is obviously also needed for continue runs.

I think I've changed my mind - could we modify `om3-scripts/payu_config/setup_scripts/setup_cice_restarts.sh` so that it only replaces the `iced.1900-01-01-10800.nc` symlink if it exists.

Then this part of the `config.yaml` can be the same across configs and so automatic cherry-picking, e.g. to remove the `setup` `userscript`, will be more likely to work.

anton-seaice: fixled :)

manifests/restart.yaml
```diff
@@ -1,3 +1,63 @@
...
+work/rpointer.ocn:
+  fullpath: /scratch/tm70/as2285/access-om3/archive/MOM6-CICE6-B/restart001/rpointer.ocn
+  hashes:
+    binhash: 9c1aa1f2ae465fd584f736c034552d1c
+    md5: f9f85cb9aad7df49a7b5f3a6fdeadccf
+work/rpointer.rof:
+  fullpath: /scratch/tm70/as2285/access-om3/archive/MOM6-CICE6-B/restart001/rpointer.rof
+  hashes:
+    binhash: 32f03c1709665b37467aa7f4451c268a
+    md5: 7b923ac9f8b4a6cd1878ef88ba85799e
```
dougiesquire, line 63: This shouldn't be here

nuopc.runconfig
```diff
@@ -288,13 +288,13 @@ CLOCK_attributes::
      lnd_cpl_dt = 99999 #not used
      ocn_cpl_dt = 1350 #ignored (coupling timestep set by nuopc.runseq) unless stop_option is nsteps
      restart_n = 1
-     restart_option = nyears
+     restart_option = ndays
```
dougiesquire, line 291: Do we really want to change this?

nuopc.runconfig
```diff
@@ -288,13 +288,13 @@ CLOCK_attributes::
...
      restart_ymd = -999
      rof_cpl_dt = 99999 #not used
      start_tod = 0
      start_ymd = 19000101
      stop_n = 1
-     stop_option = nyears
+     stop_option = ndays
```
dougiesquire, line 297: Do we really want to change this?

dougiesquire (changes requested):

anton-seaice:
Oops sorry - I fixed it. third time lucky!

dougiesquire (approved):
LGTM and seems to run successfully

PR #111, opened 2024-08-23
add comparison table to README

aekiss:
closes https://github.com/COSIMA/access-om3/issues/211


aekiss:
We could also add comparisons to configs in https://github.com/COSIMA/MOM6-CICE6-WW3, but we might need to consider a different layout as this table will quickly get too wide to display.

dougiesquire (commented):
I love this. Great idea @aekiss.

I've been trying to think of a more concise way to represent this, as we don't really need both A->B and B->A. But I've not been able to come up with anything that would be easier to use. Do you think just having a list of all the unique comparisons would make it harder to find the comparison you want?

aekiss:
I've rearranged this to a simple A->B list. I think I prefer it.

dougiesquire (approved):
Thanks @aekiss. This is really great!

PR #113, opened 2024-08-28
fix JRA55do runoff remapping bug

aekiss:
closes https://github.com/COSIMA/access-om3/issues/208

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -22,7 +22,8 @@ exe: access-om3-MOM6-CICE6
 input:
     - /g/data/vk83/experiments/inputs/access-om3/share/meshes/global.1deg/2024.01.25/access-om2-1deg-ESMFmesh.nc
     - /g/data/vk83/experiments/inputs/access-om3/share/meshes/global.1deg/2024.01.25/access-om2-1deg-nomask-ESMFmesh.nc
-    - /g/data/vk83/experiments/inputs/access-om3/share/meshes/share/2024.01.25/JRA55do-ESMFmesh.nc
+    - /g/data/vk83/configurations/inputs/access-om3/share/meshes/share/2024.08.31/JRA55do-datm-ESMFmesh.nc
+    - /g/data/vk83/configurations/inputs/access-om3/share/meshes/share/2024.08.31/JRA55do-drof-ESMFmesh.nc
     - /g/data/vk83/experiments/inputs/access-om3/share/grids/global.1deg/2020.10.22/topog.nc
     - /g/data/vk83/experiments/inputs/access-om3/mom/grids/mosaic/global.1deg/2020.05.30/ocean_hgrid.nc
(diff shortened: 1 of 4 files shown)


aekiss:
this will need to be cherry-picked into all other configs

anton-seaice:
I guess we need the same change to https://github.com/COSIMA/om3-scripts/blob/main/data_stream_xml_generation/generate_xml_datm.py ?

aekiss:
Ah yes, good point! @ezhilsabareesh8 do you have a spare moment to do that?

ezhilsabareesh8 (approved):
Thanks, @aekiss! This looks good to me, and it's running without any issues. I can change the [data_stream_xml_generation](https://github.com/COSIMA/om3-scripts/blob/main/data_stream_xml_generation) files to reflect these changes.

dougiesquire:
Sorry to be a pain in the arse, but I don't think we want to get in the habit of making changes without updating the manifest. I understand they can make cherry-picking difficult, but they're an important part of the provenance.

dougiesquire (changes requested):
Also, it would be best to regenerate the `datm.streams.xml` and `drof.streams.xml` using the new `generate_xml_*.py` in `om3-scripts` so that they have the correct provenance metadata

aekiss:
Agreed @dougiesquire, I'd forgotten these files were script-generated so we should re-generate them (and while we're at it, fix https://github.com/COSIMA/om3-scripts/issues/27 and https://github.com/COSIMA/om3-scripts/issues/29).

Re. manifests, I think we should have one commit containing only the manifest changes, and another containing everything else. That way we can just cherry-pick the 2nd commit.

dougiesquire:
> Agreed @dougiesquire, I'd forgotten these files were script-generated so we should re-generate them (and while we're at it, fix https://github.com/COSIMA/om3-scripts/issues/27 and https://github.com/COSIMA/om3-scripts/issues/29).

Sorry, I was rushing and I merged https://github.com/COSIMA/om3-scripts/pull/28 before seeing your new `om3-scripts` issues. I agree we should fix these both. @ezhilsabareesh8 could you please make these changes?

> Re. manifests, I think we should have one commit containing only the manifest changes, and another containing everything else. That way we can just cherry-pick the 2nd commit.

Sounds good to me

dougiesquire:
@aekiss, `om3-scripts` has now been updated to fix the above issues (thanks to @ezhilsabareesh8). I've updated the version of `om3-scripts` at `/g/data/vk83/apps/om3-scripts`, so you could use that to generate the new stream files.

aekiss:
Thanks @dougiesquire, I'm happy for somebody else to do that if you're waiting on it - I'm busy in a workshop tomorrow so might struggle to get to it.

aekiss:
thanks Dougie

aekiss:
Closing - PR replaced by https://github.com/ACCESS-NRI/access-om3-configs/pull/123

PR #118, opened 2024-09-02
update doc files

aekiss:
contributes to https://github.com/COSIMA/access-om3/issues/224

### DIFF
--- a/doc/ice_in
+++ b/doc/ice_in
@@ -1,29 +1,32 @@
 &setup_nml
   bfbflag = "off" 
   conserv_check = .false.
-  debug_forcing = .true.
-  debug_model = .true.
   diagfreq = 960
   dumpfreq = "y"
   dump_last = .true.
   histfreq = "d", "m", "x", "x", "x"
+  hist_suffix = "", "m", "x", "x", "x" !limited to one character due to https://github.com/COSIMA/access-om3/issues/201
+  hist_time_axis = "middle"
+  history_deflate = 1
+  history_chunksize = 180, 150
   history_precision = 8
-  ice_ic = "./input/iced.1900-01-01-10800.nc"
+  restart_deflate = 1
+  ice_ic = "./INPUT/iced.1900-01-01-10800.nc"
   lcdf64 = .false.
   npt = 35040
   pointer_file      = './rpointer.ice'
   print_global      = .false.
 /
 &grid_nml
-  bathymetry_file = "./input/topog.nc"
+  bathymetry_file = "./INPUT/topog.nc"
   grid_atm = "A"
-  grid_file = "./input/grid.nc"
+  grid_file = "./INPUT/grid.nc"
   grid_format = "nc"
   grid_ice = "B"
   grid_ocn = "A"
   grid_type = "tripole"
   kcatbound = 0
-  kmt_file = "./input/kmt.nc"
+  kmt_file = "./INPUT/kmt.nc"
   nblyr = 1
   ncat = 5
   nfsd = 1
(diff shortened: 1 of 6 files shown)


doc/ice_in
```diff
@@ -65,18 +68,18 @@
...
-  block_size_y = 15
+  block_size_x = 15
+  block_size_y = 20
   distribution_type = "cartesian"
   distribution_wght = "latitude"
   maskhalo_bound = .true.
   maskhalo_dyn = .true.
   maskhalo_remap = .true.
-  max_blocks = 10
+  max_blocks = 15
```
minghangli-uni, line 78:
```suggestion
  max_blocks = -1
```

In the latest [025deg_jra55do_ryf](https://github.com/ACCESS-NRI/access-om3-configs/tree/025deg_jra55do_ryf), `max_blocks` has been set to -1 to allow the code to automatically determine the optimal value.  
https://github.com/ACCESS-NRI/access-om3-configs/blob/f4b2d1e916c0444f4c4620382800306f4b797fe4/ice_in#L78

I can quickly update the 1deg config.

doc/ice_in
```diff
@@ -65,18 +68,18 @@
   ustar_min = 0.0005
 /
 &domain_nml
-  block_size_x = 16
-  block_size_y = 15
+  block_size_x = 15
+  block_size_y = 20
```
minghangli-uni, line 72: As was discussed in https://github.com/COSIMA/access-om3/issues/214#issuecomment-2316682306 and [cice docs](https://cice-consortium-cice.readthedocs.io/en/main/user_guide/ug_implementation.html#performance:~:text=Often%203%20to%208%20blocks%20per%20processor%20provide%20the%20decompositions%20flexiblity%20to%20create%20reasonable%20load%20balance%20configurations.), It is suggested that `max_blocks` be set in the range of 3-8 blocks per processor. 
```suggestion
  block_size_x = 30
  block_size_y = 40
```

With the updated `block_size_x` and `block_size_y`, `max_blocks` is determined to be 8, which suits the recommendation from the [cice docs](https://cice-consortium-cice.readthedocs.io/en/main/user_guide/ug_implementation.html#performance:~:text=Often%203%20to%208%20blocks%20per%20processor%20provide%20the%20decompositions%20flexiblity%20to%20create%20reasonable%20load%20balance%20configurations.).

minghangli-uni (commented):
Thanks @aekiss LGTM! Only a comment on the selection of `block_size_x`, `block_size_y`, and `max_blocks`.

doc/ice_in
anton-seaice: With the slender distributions block_size_x should equal nx_global/nprocs i.e. 360/24 = 15. Which is why these block sizes are set as they are at the moment

anton-seaice: You could try that. That would mean there is 1 block per processor, which means you couldn't eliminate any blocks which are entirely land. I am not sure how it could impact inter-process communication ? I suspect the performance will be about the same as the current configuration ?

minghangli-uni: I dont get it. If that’s the case, why does block_size_x equal nx_global / nprocs, but block_size_y is calculated differently?

anton-seaice: Its just kind of by definition, you can have multiple blocks in Y but not X. Each color is a different processor: 

Screenshot 2024-09-03 at 1 36 33 PM

aekiss:
This PR is just to update the documentation to match the config, so how about we merge this, then explore changes to the config as a separate issue? It's easy to rerun `update_doc.sh` anytime to sync the docs again.

minghangli-uni (approved):
Sure @aekiss. Feel free to merge it.

PR #124, opened 2024-09-24
Move to nri-spack

anton-seaice:
Move to nri-spack build. This is using the same access-om3 code and dependencies as the cosima-spack, although some build settings are different (e.g. the 'target' attribute). Its possible the performance is slightly different with this build.

The build still uses https://github.com/COSIMA/access-om3 for model components

PR #132, opened 2024-09-26
Move to NRI Spack Build

anton-seaice:
Companion to #124 

contributes to https://github.com/COSIMA/access-om3/issues/194 and https://github.com/COSIMA/access-om3/issues/228


minghangli-uni (approved):
LGTM! Thanks @anton-seaice

PR #134, opened 2024-09-26
Update README.md

anton-seaice:
Fixes links and contributes to https://github.com/COSIMA/access-om3/issues/221

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,12 +22,12 @@ according to the following naming scheme:
 Additional required information, like if the configuration includes
 biogeochemistry, is appended to the name.
 
-Currently the following configurations are available:
+Currently the following development configurations are available:
 
-- [`1deg_jra55do_ryf`](https://github.com/ACCESS-NRI/access-om3-configs/tree/1deg_jra55do_ryf)
-- [`1deg_jra55do_iaf`](https://github.com/ACCESS-NRI/access-om3-configs/tree/1deg_jra55do_iaf)
-- [`1deg_jra55do_ryf_wombatlite`](https://github.com/ACCESS-NRI/access-om3-configs/tree/1deg_jra55do_ryf_wombatlite)
-- [`025deg_jra55do_ryf`](https://github.com/ACCESS-NRI/access-om3-configs/tree/025deg_jra55do_ryf)
+- [`dev-1deg_jra55do_ryf`](https://github.com/ACCESS-NRI/access-om3-configs/tree/dev-1deg_jra55do_ryf)
+- [`dev-1deg_jra55do_iaf`](https://github.com/ACCESS-NRI/access-om3-configs/tree/dev-1deg_jra55do_iaf)
+- [`dev-1deg_jra55do_ryf_wombatlite`](https://github.com/ACCESS-NRI/access-om3-configs/tree/dev-1deg_jra55do_ryf_wombatlite)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,44 +43,26 @@ These configurations should **not** be used for production runs.
...
-GitHub. Detailed instructions about how to set up a fork can be found
-[here](https://docs.github.com/en/get-started/quickstart/fork-a-repo).
-
-Once you have set up your fork, we recommend cloning to a directory with a
-unique name that reflects what you wish to run. This could simply be the name of
-the configuration you plan to run, but the more detailed the name is, the less
-likely a namespace clash will happen.
-
-Finally, one needs to checkout the branch corresponding to the desired
+The first thing to do is to clone this repository. One needs to checkout the branch corresponding to the desired
```
aekiss, line 55:
```suggestion
The first thing to do is to clone this repository, then checkout the branch corresponding to the desired
```

README.md
```diff
@@ -94,6 +76,7 @@ to
 ```yaml
 runlog: true
 ```
+We recommend you create your own fork of this repository, and commit your branch to that fork. Otherwise, just committing your branch to a new github repository is a good way to track provenance and history of your work.
```
aekiss, line 79: it seems confusing to have this bit at the end of the instructions, rather than at the start - why not say to first create a fork, and then payu clone from that?

anton-seaice: This is how we've written it in other docs and is maybe our 'general' advice - @aidanheerdegen I think you have thoughts on this ?

anton-seaice: Also is creating a fork messy due to branching ? e.g. Can you fork and include one branch only ?

aidanheerdegen: > Also is creating a fork messy due to branching ? e.g. Can you fork and include one branch only ?

Yeah that was the main motivation. We're storing all the configs in a single repo for administrative convenience, but it isn't the best model for users. Additionally we have `release` and `dev` branches, which is a complication most users will not require. 

I mean they can if they want, but typically I'd expect users would want to have a repo for related experiments, which would often be variations on a single configuration.

There might be some users who'd like to have a multi-configuration repository, say if they were looking at the effect of resolution on a process, but that's quite a bit rarer, and in any case they'd want to select what configs should be in there, and not just *everything*.

aekiss: Not sure I follow. I thought we wanted users to work from a fork so they can upload their experiment branches there (for sharing, safekeeping, provenance and journal data requirements)?

As far as I can see, the only options for forking are main branch only (which is no use at all), or everything (i.e. untick "Copy the main branch only"). It's apparently only possible to have one fork per user, so it makes sense to fork every branch so users have the option of running any config in future, using one repo for everything.

anton-seaice:
Some goose confused himself with useless branch names ...

anton-seaice:
I am not sure if this PR is still relevant or will be superseeded @chrisb13 - its related to https://github.com/COSIMA/access-om3/issues/297

chrisb13:
Thanks @anton-seaice, I think the content is still highly relevant even if it ends up being in the new docs, we still need to get these issues ironed out. In general, I think it's best if we only maintain as minimal as possible set of doc's, that way we can just refer to the most up to date version rather than maintaining several sets of the same instructions.

PR #153, opened 2024-11-27
Update README.md

anton-seaice:
Add links to CI docs


chrisb13:
Looks good. Made some tiny edits around the recently inserted text. 

Would it be possible to add some clarity around these two parts:
> This Pipeline compares configurations modified in a PR (the `source` branch) against a 'ground truth' checksum, usually the `target` branch. It also verifies that commons mistakes in configurations are not made. This allows developers to know if the changes they are about to commit lead to valid and reproducible results.

Can we be clearer about what is being checked?

> The workflow requires sign off from [@ACCESS-NRI/ocean](https://github.com/orgs/ACCESS-NRI/teams/ocean) to run on Gadi.
I don't know what this means (and I'm on the team!), does it happen automatically or does the user need to make a request?

anton-seaice:
Thanks @chrisb13 - I tweaked the words a bit, not sure if they are better ?

When the manually workflow is triggered, it sends an email requesting sign off from someone in the ocean team .

anton-seaice:
@chrisb13 - I think I can only request re-review if you have clicked requested changes - can you look again please :)

chrisb13 (approved):
Looks good, thanks for the clarifying edits

PR #166, opened 2025-01-31
Use payu 1.1.6 in CI

anton-seaice:
Update to new payu for new fixes :-)

e.g. https://github.com/payu-org/payu/pull/496


CodeGat (approved):
Go for it :smile:

PR #174, opened 2025-02-17
Add Scheduled Checks Workflow

CodeGat:
I'd realised that `access-om3-configs` doesn't actually have the `schedule.yml` workflow, so scheduled checks wouldn't run. 

This PR adds scheduled monthly repro checks for `access-om3-configs`, given there are entries in `configs/ci.json`.


anton-seaice (approved):
Thanks @CodeGat 

dont forget the -wav !

PR #175, opened 2025-02-17
Turn on restart reproducability for 1deg ryf

anton-seaice:
This turns on two tests for the 1deg ryf branch.

One is that 2x1day runs give the same results as 1x2day run.
One is that 2x1days results from the same initial conditions give the same result.

Contributes to #41 

I have triggered the `checksum_slow` tests manually on gadi to confirm they run.

### DIFF
--- a/config/ci.json
+++ b/config/ci.json
@@ -3,11 +3,23 @@
     "scheduled": {
         "default": {
             "markers": "checksum"
+        },
+        "dev-1deg_jra55do_ryf": {
+            "markers": "checksum or checksum_slow"
+        },
+        "release-1deg_jra55do_ryf": {
+            "markers": "checksum or checksum_slow"
         }
     },
     "reproducibility": {
         "default": {
             "markers": "checksum"
+        },
+        "dev-1deg_jra55do_ryf": {
+            "markers": "checksum or checksum_slow"
+        },
+        "release-1deg_jra55do_ryf": {
+            "markers": "checksum or checksum_slow"
         }
     },
     "qa": {
--- a/config/ci.json
+++ b/config/ci.json
@@ -2,14 +2,10 @@
     "$schema": "https://github.com/ACCESS-NRI/schema/tree/main/au.org.access-nri/model/configuration/ci/2-0-0.json",
     "scheduled": {
         "default": {
-            "markers": "checksum"
-        },
-        "dev-1deg_jra55do_ryf": {
             "markers": "checksum or checksum_slow"
         },
-        "release-1deg_jra55do_ryf": {
-            "markers": "checksum or checksum_slow"
-        }
+        "dev-1deg_jra55do_ryf": {},
+        "release-1deg_jra55do_ryf": {}
     },
     "reproducibility": {
         "default": {


config/ci.json
```diff
@@ -3,11 +3,23 @@
     "scheduled": {
         "default": {
             "markers": "checksum"
+        },
```
jo-basevi, line 6: Are you wanting to run scheduled reproducibility checks for these branches? I think specifying branch/tag names here will trigger reproducibility tests to run every month for these branches.

anton-seaice: I thought this would mean:

all branches get the `checksum` test run monthly, and the two named branches would get `checksum` and `checksum_slow` run monthly ?

anton-seaice: > Would it be useful to have an option to run scheduled checks on all branches or all release branches?

I assumed this was what the `default` marker did? Is the `default` marker just the main branch ?

jo-basevi: `default` section stores a default config for the section. So when parsing for `markers` for example, it checks first for `marker` under the target branch/tag in the test-type section. Then if not defined, it checks for `marker` in the defaults for the test-type, and then the top-level defaults.

For `qa` and `reproducibility` sections, it's just providing test configuration for target branches. It's only the scheduled section that have the added logic of for each specific branch/tag is defined, trigger a scheduled check.

anton-seaice: Ok - I have tidied up the scheduled section based on this.

I don't have a strong view if we should be running scheduled checks on all branches - so if we just run the scheduled check on one branch thats a good step here.

jo-basevi (approved):
Changes look good to me!

config/ci.json
```diff
@@ -2,12 +2,20 @@
     "$schema": "https://github.com/ACCESS-NRI/schema/tree/main/au.org.access-nri/model/configuration/ci/2-0-0.json",
     "scheduled": {
         "default": {
-            "markers": "checksum"
-        }
+            "markers": "checksum or checksum_slow"
+        },
+        "dev-1deg_jra55do_ryf": {},
+        "release-1deg_jra55do_ryf": {}
```
dougiesquire, line 8: Does this mean these are the only branches that get scheduled checks, or are they the branches that _don't_ get scheduled checks?

I read @jo-basevi's comment and it's unclear to me

> default section stores a default config for the section. So when parsing for markers for example, it checks first for marker under the target branch/tag in the test-type section. Then if not defined, it checks for marker in the defaults for the test-type, and then the top-level defaults.

anton-seaice: I believe the first

dougiesquire: So kinda the opposite behaviour to the `qa` and `reproducibility` sections?

anton-seaice: Yeah - I think default in those sections means all branches and default in scheduled section means no branches

dougiesquire (approved):
Thanks @anton-seaice!

config/ci.json
jo-basevi: The default just holds the default options for the configuration. The logic whether to run `QA` and `reproducibility` for a branch, is in embedded in the various Github CI workflows, rather than `ci/config.yaml` file, e.g. running `reproducibility` tests when merging into a `release-*` branch. 

> Does this mean these are the only branches that get scheduled checks, or are they the branches that don't get scheduled checks?

So for scheduled checks, these are only the branches (or tags) that get scheduled checks. 

Sorry I'm not super clear, I'll look into updating the documentation for this

anton-seaice: Docs would be great @jo-basevi !

jo-basevi: Just finding what documentation currently exists, there is some notes on `model-configs-template` repository in the general [setup guide](https://github.com/ACCESS-NRI/model-configs-template?tab=readme-ov-file#in-configcijson), and in the [dev readme](https://github.com/ACCESS-NRI/model-configs-template/blob/main/README-DEV.md#ci-configuration). Do you think it would be more visible if there is a section on the `model-config-tests` repository?

anton-seaice: This is useful, two things:

- I wouldn't think to look in the `model-configs-template` repo for this? I guess we can add a link somehwere
- It doesn't capture which branches the tests are run on (which is different for schedule / repro / qa)

PR #176, opened 2025-02-17
Additional MOM6 parameter changes for 0.4.0

dougiesquire:
Changes in this PR:
- A bunch of MOM6 parameter changes - see https://github.com/COSIMA/access-om3/issues/274.
- ~Reverts to using the old CICE grid file, rather than the MOM supergrid - see https://github.com/COSIMA/access-om3/issues/275.~ - see [comment below](https://github.com/ACCESS-NRI/access-om3-configs/pull/176#issuecomment-2664173399)
- Adds up-to-date `MOM_parameter_doc.*` files.

### DIFF
--- a/docs/MOM_parameter_doc.layout
+++ b/docs/MOM_parameter_doc.layout
@@ -1,0 +1,73 @@
+! This file was written by the model and records the layout parameters used at run-time.
+
+! === module MOM ===
+GLOBAL_INDEXING = False         !   [Boolean] default = False
+                                ! If true, use a global lateral indexing convention, so that corresponding
+                                ! points on different processors have the same index. This does not work with
+                                ! static memory.
+
+! === module MOM_domains ===
+!SYMMETRIC_MEMORY_ = True       !   [Boolean]
+                                ! If defined, the velocity point data domain includes every face of the
+                                ! thickness points. In other words, some arrays are larger than others,
+                                ! depending on where they are on the staggered grid.  Also, the starting index
+                                ! of the velocity-point arrays is usually 0, not 1. This can only be set at
+                                ! compile time.
+NONBLOCKING_UPDATES = False     !   [Boolean] default = False
+                                ! If true, non-blocking halo updates may be used.
+THIN_HALO_UPDATES = True        !   [Boolean] default = True
+                                ! If true, optional arguments may be used to specify the width of the halos that
+                                ! are updated with each call.
+!STATIC_MEMORY_ = False         !   [Boolean]
+                                ! If STATIC_MEMORY_ is defined, the principle variables will have sizes that are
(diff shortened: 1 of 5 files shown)


dougiesquire:
!test repro

dougiesquire:
!test repro

dougiesquire:
!test repro

dougiesquire:
[@anton-seaice would like to keep the CICE supergrid functionality in this configuration](https://github.com/COSIMA/access-om3/issues/275#issuecomment-2664162564), so I'm going to force push without that commit.

dougiesquire:
@minghangli-uni would you be able to review?

docs/MOM_parameter_doc.layout
```diff
@@ -0,0 +1,69 @@
...
+                                ! If true, non-blocking halo updates may be used.
+THIN_HALO_UPDATES = True        !   [Boolean] default = True
+                                ! If true, optional arguments may be used to specify the width of the halos that
+                                ! are updated with each call.
+!STATIC_MEMORY_ = False         !   [Boolean]
+                                ! If STATIC_MEMORY_ is defined, the principle variables will have sizes that are
+                                ! statically determined at compile time.  Otherwise the sizes are not determined
+                                ! until run time. The STATIC option is substantially faster, but does not allow
+                                ! the PE count to be changed at run time.  This can only be set at compile time.
+AUTO_MASKTABLE = False          !   [Boolean] default = False
```
minghangli-uni, line 26:
```suggestion
AUTO_MASKTABLE = True          !   [Boolean] default = False
```

minghangli-uni (commented):
Thanks @dougiesquire. Just one comment: Are we planning to enable `AUTO_MASKTABLE` for the 1deg configuration? We did this for the 0.25deg config https://github.com/ACCESS-NRI/access-om3-configs/pull/85

docs/MOM_parameter_doc.layout
dougiesquire: Yeah, I don't see why not. But I'll just do some quick checks to make sure things run and don't slow down

dougiesquire: Basic tests indicate no change in performance. It does change answers.

anton-seaice: I would guess at this resolution there are very few blocks that get removed ?

Saying that - probably just turn it on in all configs for consistency

minghangli-uni: For the 1deg with the current processor layout, only 22 land blocks are eliminated, so with a short run we cannot tell the runtime performance difference.

anton-seaice:
!test repro commit

PR #177, opened 2025-02-18
Demo restart repro through CI

anton-seaice:


anton-seaice:
!test repro

anton-seaice:
!test repro

anton-seaice:
I was trying to show here that restart repro was broken. However it's not very clear because the checksums for historical repro are out of date - we'll update them in #180 and come back to this

dougiesquire:
I'll update the checksums with the "Generate Initial Checksums" workflow since it would be handy to have this demo for the OSIT meeting

anton-seaice:
!test repro

anton-seaice:
!test repro

PR #181, opened 2025-02-19
Catch up 1 deg configs with `dev-025deg_jra55do_ryf`

dougiesquire:
This PR implements a few changes that were made to the `dev-025deg_jra55do_ryf` config but were not included in the 1 deg configs:

 - Turn off tidal forcing. See https://github.com/COSIMA/access-om3/issues/276.
 - Use `DIABATIC_FIRST = False` (default). See https://github.com/COSIMA/access-om3/issues/137.
 - Revert changes to use the MOM supergrid in CICE. See https://github.com/COSIMA/access-om3/issues/275.
 - Use MOM `RESTART_CONTROL = 1` (default). With the `CESMCOUPLED` cpp we don't actually use this parameter.
 - Use MOM `MINIMUM_DEPTH = 0.0` (default). See https://forum.access-hive.org.au/t/namelist-configuration-discussion-meeting/1917/9?u=dougiesquire.
 - Use default `BAD_VAL_*` values. This runs successfully for 1 year.
 - Use `HORIZ_VARYING_BACKGROUND = False`; `KD = 0.0` (default). See https://github.com/COSIMA/access-om3/issues/199
 - Default to year long runs


dougiesquire:
!test repro

dougiesquire:
!test repro

dougiesquire:
Note the reproducibility tests reported in the commit above failed because MOM aborted due to velocity truncations

dougiesquire:
!test repro

dougiesquire:
!test repro

dougiesquire:
!test repro

dougiesquire:
!test repro

dougiesquire:
I'm a bit surprised that setting `MINIMUM_DEPTH = 0.0` changed answers (very slightly) since the minimum depth of `TOPO_FILE` is 11.8 m...

Screenshot 2025-02-20 at 12 09 32 pm

dougiesquire:
@minghangli-uni, would you be able to take a look at this?

minghangli-uni (commented):
Thanks @dougiesquire. I was able to run it successfully for a month. Just one question: while `DIABATIC_FIRST` is agreed to be False, another issue is for DT and DT_THERM. I haven’t investigated this for the 1deg config yet, but I believe that the coupling timestep and cice therm timestep should at least match the MOM DT. This ensures that both components exchange fluxes at the same time, regardless of the final choice for dt_therm. This is what we do for the 0.25deg at the moment. So DT=DT_CPL=3600s? 1800s would be a bit low for the 1deg config.

dougiesquire:
Thanks @minghangli-uni. I'm not sure I understand all of your point sorry. The coupling timestep in this configuration is 3600s and both MOM and CICE exchange fluxes at that timestep, right? With `DT = 1800`, MOM just does two dynamic timesteps for every coupling (and thermodynamic) timestep.

I agree with you that the MOM dynamic timestep should probably be increased. But I think this probably requires some basic testing and should probably be done in a subsequent PR. @anton-seaice already opened an issue about this last week - see https://github.com/COSIMA/access-om3/issues/268

minghangli-uni:
My thinking is if dt is half of dt_cpl, would this delay feedback between ocean and cice? E.g., fluxes remain constant for two ocean baroclinic steps before being updated.

dougiesquire:
I tried running with `DT = 3600.0`. I got a lot of warnings like this:
```
WARNING from PE     3: btstep: eta has dropped below bathyT:  -1.4331200363952876E+01 vs.  -1.1805749893188477E+01 at  -1.4450E+02 -7.5568E+01    136     10
```
and then MOM aborted due to velocity truncations.

dougiesquire:
!test repro commit

dougiesquire:
!cherry-pick 0707fa2af73bef36cc1c19b7db75a9e36e6f081d 76dd7cb7698990314eb4b5225feaa71a90e9272e ee678b95e197a2951eb9c272999c48474c4af3b5 63f98c2628e32761c14449df203b96980ef9c146 11e544d5a4ea50eeeb7f298c15dbb441bb638b80 6d239ca13f3fb3e2449529c4ada332a632c35b60 c0aea88afa6e3da5a7a297c30dc3e20522ce3c65 into dev-1deg_jra55do_iaf

dougiesquire:
Duh!

dougiesquire:
!cherry-pick b7947cf1123d784c942547300d6fc6c8a4674631 d39270f261178d993dacbc125cf39dfaf0822cf8 2d0a6cdcdf06d885c623c384d109bbce713f664c 903774e71af678272ed8173bae5d8eb1b41a1b60 8a2fcab4c0d5e6a9a8dfabb5dd698af2061d1c7c f711d80fd8109aa85d83d01e228a526e700e8b60 d0278aa85adfe6f338e677e60cf20b06526c1a70 into dev-1deg_jra55do_iaf

Issue #193, opened 2025-02-25
Update CI config


anton-seaice:
There are several useful changes implemented / underway which we should update `config.yaml` to use:

https://github.com/ACCESS-NRI/model-config-tests/pull/111 : allows using regex matching in branch names, e.g.

```
"qa": {
        "dev-.*": { 
            "markers": "dev_config"
        },
        "default": {
           "markers": "config or dev_config"
       }
   },
```

https://github.com/ACCESS-NRI/model-config-tests/pull/113 : allows using the main branch of `model-config-tests`. We should set dev branches to use main branch and release branches to use a release version

https://github.com/ACCESS-NRI/model-config-tests/pull/110 : fixes restart repro

https://github.com/ACCESS-NRI/model-config-tests/pull/103 : added the dev_config marker

anton-seaice:
Reopening so we can fix the regex syntax - per https://github.com/ACCESS-NRI/model-config-tests/issues/118#issue-2882958102

jo-basevi:
Now with https://github.com/ACCESS-NRI/model-config-tests/pull/126 which prevents running scheduled tests on any branch/tag that does not exist in the repository, it should be possible to include regex under `scheduled` tests. Note that regex currently only be used for getting common testing configuration such as `payu-version` - it won't trigger scheduled tests for every release branch. So for the following `config/ci.json`:
```json
"scheduled": {
        "default": {
            "markers": "checksum or checksum_slow or access_om3 or config"
        },
        "dev-1deg_jra55do_ryf": {
            "markers": "checksum or checksum_slow or access_om3 or dev_config"
        },
        "release-1deg_jra55do_ryf": {},
        "release-.*": {
            "model-config-tests-version": "0.0.12",
            "payu-version": "1.1.6"
        }
#...
```
The only scheduled test will run would be `dev-1deg_jra55do_ryf`. When `release-1deg_jra55do_ryf` exists in the repository, it will also trigger a scheduled test, and it will fall back to `release-.*` and `default` for the pytest configuration.

anton-seaice:
> When `release-1deg_jra55do_ryf` exists in the repository

If we add this to the ci.json now, will this test fail? Or not run ?

jo-basevi:
It will not run a scheduled test for `release-1deg_jra55do_ryf`

PR #202, opened 2025-03-03
Update ci.json

anton-seaice:
Fix the regex for scheduled tests per #193

### DIFF
--- a/config/ci.json
+++ b/config/ci.json
@@ -7,7 +7,7 @@
         "dev-1deg_jra55do_ryf": {
             "markers": "checksum or checksum_slow or access_om3 or dev_config"
         },
-        "release-.*": {
+        "release-1deg_jra55do_ryf": {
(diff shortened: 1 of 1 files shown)


config/ci.json
```diff
@@ -7,7 +7,7 @@
         "dev-1deg_jra55do_ryf": {
             "markers": "checksum or checksum_slow or access_om3 or dev_config"
         },
-        "release-.*": {
+        "release-1deg_jra55do_ryf": {
```
jo-basevi, line 10: While there isn't a `release-1deg_jra55do_ryf` branch this will also fail to run so I think it'll be best to add this back in when this branch exists.

anton-seaice: Thanks!

jo-basevi (approved):
Thanks @anton-seaice, looks good to me

PR #482, opened 2025-03-28
Add test-markers input

CodeGat:
References ACCESS-NRI/model-config-tests#147

## Background

The 'generate initial checksums' workflow used to assume that the pytest markers were always `checksum`. The markers had changed, which led to the workflow erroneously succeeding without testing anything. 

Instead, make the pytest markers for initial checksums settable by the user, giving a sensible default of `repro` rather than hardcoding it. 

## The PR

* Add a new, not required input of `test-markers` that defaults to `repro`. 
* Log this output
* Pass it to the `model-config-tests` workflow entrypoint.

### DIFF
--- a/.github/workflows/generate-initial-checksums.yml
+++ b/.github/workflows/generate-initial-checksums.yml
@@ -19,6 +19,11 @@ on:
         type: string
         required: false
         description: "If checksums are being committed: The initial version for the git tag associated with the committed checksums."
+      test-markers:
+        type: string
+        required: false
+        default: repro_historical
+        description: Test markers to generate checksums, in the form of pytest test marker strings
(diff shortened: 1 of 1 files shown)


jo-basevi (approved):

jo-basevi:
Just noting that logic in the caller workflow (`generate initial checksums.yml`), has been moved now to `access-nri/model-config-tests/.github/workflows/generate-checksums.yml` in https://github.com/ACCESS-NRI/model-config-tests/pull/147. So this PR now also:
- Moves config and log-inputs jobs into the referenced workflow
- Adds a new `committed-checksum-create-release` input to optionally create releases

PR #520, opened 2025-04-10
davide/docs setup

atteggiani:
- **Added config files**
- **Added test page**
- **first attempt at porting wiki from https://github.com/cosima/access-om3/wiki to the new mkdoc system**
- **first go at trying github action to build docs**
- **Davide/docs setup (#517)**
- **Restructured mkdocs.yml and added needed files**


atteggiani:
> Or (ideally) can CMEPS-coupled configurations still be the first sub-section in Configurations but that it lives on the main branch.

Yes that can be the "overview" page when the users clicks on Configurations (or CMEPS Configurations), and it will be in the main branch. 

> I think there will need to be some link updates soon-ish
All the content updates can be managed with later PRs.


I will go on and merge this for now, so you can start working on the "main" documentation updates.

PR #549, opened 2025-05-07
Use 2025.05.000/1 build using oneapi compiler

anton-seaice:
**1. Summary**:

What has changed?

Moved to the latest build, removed an unused config line and updated degrees to km in the input files. Updated the run sequence to be more efficient, turned on cice grid output.

Why was this done?

This moves to a oneapi build, and spack packages build seperately for each model component

**2. Issues Addressed:**

- See linked issues in commits

**3. Depedencies (e.g. on payu, model or om3-scripts)**

This change requires changes to (note required version where true):
- [ ] payu:
- [x] access-om3:
- [ ] om3-scripts:

**4. Ad-hoc Testing**

A 12-month run before and after the change most global averages look unchanged (e.g.) :

image

With the exception of salt added in the ocean_scalar file:

image

But I think that's dodgy accounting / hitting the precision limit of double vars, as comparing the other two salt flux fields gives a similar figure 
Screenshot 2025-05-07 at 6 01 47 pm

Restarts at the end of the 12month run show chaotic differences (intel minus oneapi, as a fraction):

image

Except for some patterning in surface salinity!:

Amazon River December average:

Screenshot 2025-05-07 at 6 16 53 pm

Red Sea December average

Screenshot 2025-05-07 at 6 17 13 pm

Large number of scrappy plots are [here](https://github.com/anton-seaice/sandbox/blob/main/intelvsoneapi.ipynb) - can be tidied on request

**5. CI Testing**

- [x] `!test repro` has been run

**6. Reproducibility**

- [ ] Yes
- [x] No - `!test repro commit` has been run.  

**7. Documentation**

- [ ] Yes
- [x] N/A

- [ ] Yes: 
- [x] N/A

**8. Formatting**

- [ ] Yes
- [x] N/A

**9. Merge Strategy**

- [x] Merge commit
- [ ] Rebase and merge
- [ ] Squash

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -46,7 +46,7 @@ modules:
     use:
         - /g/data/vk83/modules
     load:
-        - access-om3/2025.01.2
+        - access-om3/2025.05.000
         - nco/5.0.5
 
 payu_minimum_version: 1.1.6
--- a/config.yaml
+++ b/config.yaml
@@ -46,7 +46,7 @@ modules:
     use:
         - /g/data/vk83/modules
     load:
-        - access-om3/2025.05.000
+        - access-om3/2025.05.001
         - nco/5.0.5
 
 payu_minimum_version: 1.1.6
--- a/config.yaml
+++ b/config.yaml
@@ -14,23 +14,23 @@ jobfs: 10GB
 mem: 960GB
 
 walltime: 02:00:00
-jobname: 1deg_jra55do_ryf
+jobname: 100km_jra55do_ryf
 
 model: access-om3
 
 exe: access-om3-MOM6-CICE6
 input:
-    - /g/data/vk83/configurations/inputs/access-om3/share/meshes/global.1deg/2024.01.25/access-om2-1deg-ESMFmesh.nc
-    - /g/data/vk83/configurations/inputs/access-om3/share/meshes/global.1deg/2024.01.25/access-om2-1deg-nomask-ESMFmesh.nc
+    - /g/data/vk83/configurations/inputs/access-om3/share/meshes/global.100km/2024.01.25/access-om2-100km-ESMFmesh.nc
+    - /g/data/vk83/configurations/inputs/access-om3/share/meshes/global.100km/2024.01.25/access-om2-100km-nomask-ESMFmesh.nc
     - /g/data/vk83/configurations/inputs/access-om3/share/meshes/share/2024.09.16/JRA55do-datm-ESMFmesh.nc
     - /g/data/vk83/configurations/inputs/access-om3/share/meshes/share/2024.09.16/JRA55do-drof-ESMFmesh.nc
-    - /g/data/vk83/configurations/inputs/access-om3/share/grids/global.1deg/2020.10.22/topog.nc
-    - /g/data/vk83/configurations/inputs/access-om3/mom/grids/mosaic/global.1deg/2020.05.30/ocean_hgrid.nc
-    - /g/data/vk83/prerelease/configurations/inputs/access-om3/mom/grids/vertical/global.025deg/2025.03.12/ocean_vgrid.nc
-    - /g/data/vk83/configurations/inputs/access-om3/mom/initial_conditions/global.1deg/2020.10.22/ocean_temp_salt.res.nc
-    - /g/data/vk83/configurations/inputs/access-om3/mom/surface_salt_restoring/global.1deg/2020.05.30/salt_sfc_restore.nc
-    - /g/data/vk83/configurations/inputs/access-om3/cice/grids/global.1deg/2024.05.14/kmt.nc
-    - /g/data/vk83/configurations/inputs/access-om3/cice/initial_conditions/global.1deg/2023.07.28/iced.1900-01-01-10800.nc
+    - /g/data/vk83/configurations/inputs/access-om3/share/grids/global.100km/2020.10.22/topog.nc
+    - /g/data/vk83/configurations/inputs/access-om3/mom/grids/mosaic/global.100km/2020.05.30/ocean_hgrid.nc
+    - /g/data/vk83/prerelease/configurations/inputs/access-om3/mom/grids/vertical/share/2025.03.12/ocean_vgrid.nc
+    - /g/data/vk83/configurations/inputs/access-om3/mom/initial_conditions/global.100km/2020.10.22/ocean_temp_salt.res.nc
+    - /g/data/vk83/configurations/inputs/access-om3/mom/surface_salt_restoring/global.100km/2020.05.30/salt_sfc_restore.nc
+    - /g/data/vk83/configurations/inputs/access-om3/cice/grids/global.100km/2024.05.14/kmt.nc
+    - /g/data/vk83/configurations/inputs/access-om3/cice/initial_conditions/global.100km/2023.07.28/iced.1900-01-01-10800.nc
     - /g/data/vk83/configurations/inputs/JRA-55/RYF/v1-4/data
  
 collate: false
--- a/datm_in
+++ b/datm_in
@@ -8,8 +8,8 @@
   flds_preso3 = .false.
   flds_wiso = .false.
   iradsw = 1
-  model_maskfile = "./INPUT/access-om2-1deg-nomask-ESMFmesh.nc"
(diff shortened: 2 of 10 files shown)


anton-seaice:
!test repro commit

anton-seaice:
!test repro commit

anton-seaice:
!test repro commit

anton-seaice:
!test repro commit

anton-seaice:
I think this is good to go. The salinity differences are interesting, but neither is clearly better than the other and we might need to work on some runoff spreading anyway.

dougiesquire:
Thanks for the thorough checks @anton-seaice. Naively, I'm surprised that there are such systematic differences (seemingly related to salinity restoring). @manodeep, @micaeljtoliveira, is this sort of thing expected?

manodeep:
Hard to say for sure - for example, changes on the order of $10^{-12}$ is possible, but ideally should be at the $10^{-15}-10^{-16}$ range (`DBL_EPSILON`). Changes to the temperature of $\pm 0.1$ seems high at first sight - are the temperatures a result of some accumulation (eg. sums)?. 

Related, are the oneAPI compiler options mimic'ing the ifort compiler options?

anton-seaice:
Build flags are unchanged from the intel builds:

For example the main fortran ones in CICE they are:

`
  set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -qno-opt-dynamic-align  -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model precise") 
`

I only really worried about flags in terms of "are the results deterministic" rather than this context.

Based on previous discussions
`-qno-opt-dynamic-align` is [supported](https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2025-0/qopt-dynamic-align-qopt-dynamic-align.html)
`-convert big_endian ` and `-assume byterecl ` are confusing and not available in oneapi, but it also seems like they may not do alot currently
` -assume realloc_lhs` is no longer needed 


MOM6 has `-r8` set - whats the oneapi equivalent for that ?





> Hard to say for sure - for example, changes on the order of 10 − 12 is possible, but ideally should be at the 10 − 15 − 10 − 16 range (`DBL_EPSILON`). Changes to the temperature of ± 0.1 seems high at first sight - are the temperatures a result of some accumulation (eg. sums)?.

I think this would be true after say one calculation - maybe one timestep in this case? But after a year the random changes would have moved chaotically through the system

anton-seaice:
On second read, MOM defines saltflux as a real without specifying precision (e.g. https://github.com/ACCESS-NRI/MOM6/blob/e088c8b7f6c2b18b72edd568aa009e13396ec0c3/config_src/drivers/nuopc_cap/mom_surface_forcing_nuopc.F90#L166).

`
real, pointer, dimension(:,:) :: salt_flux  
`


 So maybe setting -r8 converted that to type double in intel-classic and oneapi does not?

micaeljtoliveira:
> So maybe setting -r8 converted that to type double in intel-classic and oneapi does not?

`-r8` is supported by oneapi. See "Alternate Options" at the bottom of [this page](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2025-1/real-size.html).

anton-seaice:
I confirmed that using the new-build system with the old compiler give [the same answers](https://github.com/ACCESS-NRI/access-om3-configs/pull/554) as the old build system, and that using the old-build system with oneapi gives [the same answers as this PR](https://github.com/ACCESS-NRI/access-om3-configs/pull/553).

So its definitely the change in optimisation from oneapi compared to intel-classic which gives the differences in the first plot.

manodeep:
@anton-seaice Assuming you have added the `-r8`, nothing in the compiler flags that you have mentioned in [this comment](https://github.com/ACCESS-NRI/access-om3-configs/pull/549#issuecomment-2861880441) jump out as being extra aggressive. Which brings me back to my original comment about the magnitude of the deviation - 0.1 seems fairly large for temperature (which I assume ranges ~ [-100 deg C, +60 deg C]). May be, as you pointed out, the duration of 1 year is at play; however, if the errors were stochastic, I would expect some to cancel out, and hence accumulate at a lower rate than what seems to be happening (ie, 14 orders of mag. larger than the expected epsilon of $10^{-15}$ for a single calculation)

I would also recommend adding the `-assume nan_compares -assume ieee_compares`. There is a comment in the [ifort->ifx porting guide](https://www.intel.com/content/www/us/en/developer/articles/guide/porting-guide-for-ifort-to-ifx.html) which should only apply to `-fp-model=fast` but may be better to be defensive, especially in light of (significantly?) changed results.

MartinDix:
I'm not at all surprised by the magnitude of the December temperature differences.

It'd be interesting to see a run with a tiny initial random temperature perturbation to check if the salinity differences are larger than expected from roundoff.

manodeep:
@MartinDix With a drift of 0.1 deg/year, wouldn't the net drift over ~100 years become quite unphysical? Or is the expectation that the drift will level off sooner rather than later?

From the numerical side, here's what I was thinking - each temperature update depends on ~100 neighbouring cells. With ~hour timesteps, there are  about 10k updates in a year for a cell - so a total of 10^6 floating point ops go into the temperature updates per year. There's still about 7-8 orders of magnitude difference to go from $10^{-15}$ to the 0.1 deg drift from classic intel, so I must be missing some additional physics/numerical context. Or is this just the consequence of chaotic systems?

MartinDix:
I doubt there's a net drift. The largest temperature changes look like shifts in the sea ice edge and in the Gulf Stream position. Numerical differences tend to grow exponentially with time rather than linearly because changes in winds/currents affect advection as well.

anton-seaice:
> Numerical differences tend to grow exponentially with time rather than linearly because changes in winds/currents affect advection as well.

What is going on with the runoff/sos field too then might be similar

anton-seaice:
> With a drift of 0.1 deg/year, wouldn't the net drift over ~100 years become quite unphysical?

I would assume this is random variation rather than a drift but could check will run Martin's suggested test to see how different it looks.

manodeep:
Thanks Martin! Yes, if the maximum difference is 0.1 deg, then that's a better case than a global 0.1 deg drift in the temperatures.

MartinDix:
From a perturbed ACCESS-CM2 ensemble, RMS differences relative to the ensemble mean
Picture1

Ocean only differences should grow a lot more slowly because of the atmospheric forcing.

anton-seaice:
last call for thoughts @MartinDix @aekiss @chrisb13 ?

I'll plan to merge tomorrow

dougiesquire:
> I'll plan to merge tomorrow

Is it worth merging given we're about to do a new release? Will you update all configs?

anton-seaice:
I asked you this yesterday and you said to merge it :)

I don't mind, happy to update to the new build

dougiesquire:
Did I? Sorry, I either misunderstood or I'm completely inconsistent. Both are definitely possible.

anton-seaice:
I think updating to the new build is sensible

minghangli-uni:
!test repro

anton-seaice:
!test repro commit

anton-seaice:
It looks like 2025.05.001 is answer changing compared to 2025.05.000. I guess that's related to adding optimisation flags ?

anton-seaice:
@dougiesquire - this is ready for review

dougiesquire:
> It looks like 2025.05.001 is answer changing compared to 2025.05.000. I guess that's related to adding optimisation flags ?

You also changed the radius of the earth in access3-share

anton-seaice:
Is something broken in MOM? the docs changed

dougiesquire:
> Is something broken in MOM? the docs changed

Are you sure you didn't start from a restart?

anton-seaice:
Ah I did! I see!

anton-seaice:
Ok - I actually think it's ready for review this time

anton-seaice:
Do you have a merge button @dougiesquire (use a merge commit) - it seems to be waiting for a test result which passed? I wonder if that was related to me force pushing

dougiesquire:
I've tried kicking off the checks again. If that fails, I'll bypass and merge

PR #571, opened 2025-05-20
Remove Leith related and use MOM symmetry

minghangli-uni:
Restart repro check using MOM symmetry memory and removing Leith related parameters in MOM config

### DIFF
--- a/MOM_input
+++ b/MOM_input
@@ -396,7 +396,7 @@ LAPLACIAN = True                !   [Boolean] default = False
                                 ! If true, use a Laplacian horizontal viscosity.
 AH = 1.0E+12                    !   [m4 s-1] default = 0.0
                                 ! The background biharmonic horizontal viscosity.
-LEITH_AH = True                 !   [Boolean] default = False
+!LEITH_AH = True                 !   [Boolean] default = False
                                 ! If true, use a biharmonic Leith nonlinear eddy viscosity.
 BACKSCATTER_UNDERBOUND = False  !   [Boolean] default = True
                                 ! If true, the bounds on the biharmonic viscosity are allowed to increase where
@@ -405,7 +405,7 @@ BACKSCATTER_UNDERBOUND = False  !   [Boolean] default = True
                                 ! when no Laplacian viscosity is applied.  The default is true for historical
                                 ! reasons, but this option probably should not be used because it can contribute
                                 ! to numerical instabilities.
-LEITH_BI_CONST = 128.0          !   [nondim] default = 0.0
+!LEITH_BI_CONST = 128.0          !   [nondim] default = 0.0
                                 ! The nondimensional biharmonic Leith constant, typical values are thus far
                                 ! undetermined.
 FRICTWORK_BUG = False           !   [Boolean] default = True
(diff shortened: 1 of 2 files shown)


minghangli-uni:
!test repro

minghangli-uni:
`test_repro_restart (test-venv.lib.python3.10.site-packages.model_config_tests.test_bit_reproducibility.TestBitReproducibility) failed`

dougiesquire:
The `repro_restart` failure is a false positive (false negative?):

```diff
git diff --no-index restart-1d-1-checksum.json restart-2d-0-checksum.json
diff --git a/restart-1d-1-checksum.json b/restart-2d-0-checksum.json
index 845c360..1af0d9b 100644
--- a/restart-1d-1-checksum.json
+++ b/restart-2d-0-checksum.json
@@ -77,7 +77,7 @@
       "60BE01A31CD24290"
     ],
     "v": [
-      "224085FA0A1B5A05"
+      "A24085FA0A1B5A05"
     ],
     "v2": [
       "29EAC3FCCE83E191"
 ```

**Comment by minghangli-uni** [comment]
!test repro

**Comment by minghangli-uni** [comment]
```diff
$ git diff --no-index restart-1d-1-checksum.json restart-2d-0-checksum.json
diff --git a/restart-1d-1-checksum.json b/restart-2d-0-checksum.json
index 845c360..1af0d9b 100644
--- a/restart-1d-1-checksum.json
+++ b/restart-2d-0-checksum.json
@@ -77,7 +77,7 @@
       "60BE01A31CD24290"
     ],
     "v": [
-      "224085FA0A1B5A05"
+      "A24085FA0A1B5A05"
     ],
     "v2": [
       "29EAC3FCCE83E191"
```

This is expected to have the false negative again. But I would expect the checksum to differ from the one in https://github.com/ACCESS-NRI/access-om3-configs/pull/571#issuecomment-2892740801, since a parameter has been changed?

dougiesquire:
> But I would expect the checksum to differ from the one in https://github.com/ACCESS-NRI/access-om3-configs/pull/571#issuecomment-2892740801, since a parameter has been changed?

You added `THERMO_SPANS_COUPLING = True` but the thermo timestep doesn't span coupling, so I guess this does nothing?

minghangli-uni:
```
DT = 1800.0                     !   [s]
                                ! The (baroclinic) dynamics time step.  The time-step that is actually used will
                                ! be an integer fraction of the forcing time-step (DT_FORCING in ocean-only mode
                                ! or the coupling timestep in coupled mode.)
DT_THERM = 3600.0               !   [s] default = 1800.0
                                ! The thermodynamic and tracer advection time step. Ideally DT_THERM should be
                                ! an integer multiple of DT and less than the forcing or coupling time-step,
                                ! unless THERMO_SPANS_COUPLING is true, in which case DT_THERM can be an integer
                                ! multiple of the coupling timestep.  By default DT_THERM is set to DT.
```
DT=1800s and DT_THERM is 3600s, so I am expecting something would change with THERMO_SPANS_COUPLING on?

dougiesquire:
Isn't the coupling timestep 3600s?

minghangli-uni:
closing it as completed

PR #588, opened 2025-06-04
547 dev mc 100km jra ryf+wombatlite

anton-seaice:
companion change to #549


anton-seaice:
!test repro commit

anton-seaice:
```
FATAL from PE     3: g_tracer_print_info: caco3bury : src_file is not set in the field_table
```

dougiesquire (approved):
With the WOMBATlite version update in this PR, this configuration is now 2.5x slower than [`dev-MC_100km_jra_ryf`](https://github.com/ACCESS-NRI/access-om3-configs/tree/dev-MC_100km_jra_ryf%2Bwombatlite) (3 hrs/yr vs 1.2 hrs/yr). Need to load balance and explore extending the tracer time step.

minghangli-uni (approved):
> Need to load balance and explore extending the tracer time step.

It would be good to explore the tracer timestep first using the old nuopc.runseq, then I can help with the load balancing using the updated runsequence.

dougiesquire:
> I’ve noticed that all diagnostics related to `generic_wombatlite_d2/generic_wombatlite` are missing from `docs/available_diags.000000`, even though they are listed in the `diag_table`. Was this intentional?

Yup thanks, this is due to this issue: https://github.com/ACCESS-NRI/GFDL-generic-tracers/issues/29

WOMBAT diagnostics are currently handled directly by FMS, rather than through MOM6

dougiesquire:
!test repro commit

dougiesquire:
!test repro commit

minghangli-uni (dismissed):

minghangli-uni (approved):
LGTM but I haven’t checked the updated wombatlite diags.

dougiesquire:
Thanks. The new diagnostics are just taken from the current ESM1.6 diagnostic profile. I imagine we'll refine them as we start to look in detail at the output, but this should be a good starting point.

PR #646, opened 2025-07-16
Stub for upcoming release-MC_25km_jra_ryf beta1.0 release

chrisb13:
Test to see if we can build the om3 config docs with a stub in it.

### DIFF
--- a/documentation/stub/release-MC_25km_jra_ryf.md
+++ b/documentation/stub/release-MC_25km_jra_ryf.md
@@ -1,0 +1,3 @@
+## release-MC_25km_jra_ryf
+
+This is a test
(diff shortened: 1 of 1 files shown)


dougiesquire:
Yeah, @chrisb13's branch name here follows the conventions for branch protections (`dev-*`, or `release-*`). @chrisb13, in general we try to follow `{issue_number}-{configuration_branch}` for feature branches - see [here](https://access-om3-configs.access-hive.org.au/pages/Git-practices/)

atteggiani:
> @atteggiani, do you have a suggestion for a development branch name, for a stub, such that it would be previewable? Perhaps having "stub" in the branch title could help? (e.g. in this PR it was `dev-MC_25km_jra_ryf_CBSTUBtest`)

I would only suggest a name that follows a well-defined pattern, because that pattern is needed in the [preview_website.pattern](https://github.com/ACCESS-NRI/mkdocs_include_configuration_stubs_plugin?tab=readme-ov-file#options) option of the `mkdocs_include_configuration_stubs_plugin` to define which branches/tags should be included in the search for the stubs to be included in the preview.

PR #692, opened 2025-08-07
Add data_table with fields for co2 and o2 fluxes

dougiesquire:
This PR adds a `data_table` with fields for co2 and o2 fluxes.

@helenmacdonald I recall Lizzie has already created the dust input on the correct grid? Do you have access to Lizzie’s dust file and if so can you add it?

You'll need to add the path to the file to the `config.yaml` and the following to the `data_table`:

```
"OCN", "dry_dep_fe_flux_ice_ocn", "dust", "./INPUT/dust.nc", "none", -1.0e-06
```

If it's not ready to go, we could just merge this and add dust forcing later

### DIFF
--- a/data_table
+++ b/data_table
@@ -1,0 +1,7 @@
+# MOM/FMS data table
+#
+# gridname | fieldname_code           | fieldname_file | file_name | ongrid  | factor
+# ----------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


helenmacdonald (approved):

dougiesquire:
Sounds good. That should be an easy fix. If you can point me at the file I can take a look

PR #696, opened 2025-08-07
making it run faster 

helenmacdonald:
DT and DT_THERM has been increased to 300 and 2400 from 50 and 300 to speed things up. I did a 30 day run test with no instabilities.
AUTO_MASKTABLE has been set to True.


pearseb (approved):
Looks good. Thanks Helen!

helenmacdonald:
> Usually we'd also update the files in the `docs` directory when `MOM_input` is modified. These are automatically updated at the end of a Payu run.
> 
> Given this isn't a supported config yet, I'm happy to let it slide

Thanks Dougie - good to know. I will make sure to do it next time.

PR #699, opened 2025-08-07
Add stub for MOM6 and CICE parameter choices documentation

minghangli-uni:
Add stubs for mom6 and cice docs


chrisb13 (dismissed):
Looks good almost but two changes

needs to be in `documentation/stubs`
needs to be named `dev-MC_25km_jra_ryf.md`

chrisb13 (approved):

PR #701, opened 2025-08-08
Fixing units, updating documentation and pointing to a new file

helenmacdonald:
- The docs have been updated (they have not been updated in a while so some of the changes you see are not recent ones)
- config.yaml is pointing to an updated bathymetry file that swaps the fill value from nans to 0 (to help auto masking find the land). There is an associated update in manifests
- field table has had input units set to `none` to prevent scaling at runtime


pearseb (approved):
Group approval from the team yay

PR #694, opened 2025-08-07
(Issue 691) OM3 config doc splurge Aug 2025 PR

chrisb13:
Executing what's described [here](https://github.com/ACCESS-NRI/access-om3-configs/issues/691#issuecomment-3161916957).

--

~~DRAFT PR, do NOT review yet please!~~

PR #715, opened 2025-08-13
Hide the `pages` url segment from the website

atteggiani:
## Overview
In mkdocs, the URL for a page is defined by the path (relative to the [docs_dir](https://www.mkdocs.org/user-guide/configuration/#docs_dir)) of the file the page is rendered from.

As a default, this would make a page being rendered from the file in `<docs_dir>/example/path/url/page.md` have its URL as `<website_root_url>/example/path/url/page`.

In this repo, the documentation pages within the `docs_dir` (`docs` folder) are further stored within the `pages` directory. This can be useful to better separate them from other files that don't constitute "pages" for the website. However, this makes the URL of a page be `<website_root_url>/pages/...`.
The `pages` url segment is not very beneficial and could be removed for a better readability and searchability within the website.

This PR adds a simple mkdocs hook to "hide" the `pages` segment from the website url, still maintaining the directory tree structure of the repo.


atteggiani:
> Is the concept that we would merge this after [this](https://github.com/ACCESS-NRI/access-om3-configs/pull/694#issuecomment-3182719250)? Or perhaps it doesn't matter?

It doesn't really matter

chrisb13 (approved):
Thanks @atteggiani !

Issue #447, opened 2025-02-09
Reproducibility fails with 0.4.0 build and the latest configuration


minghangli-uni:
Below are the 5-day `ocean.stats` results highlighting the repro issue (an issue is raised here https://github.com/COSIMA/access-om3/issues/265) when comparing the 0.3.0 and 0.4.0 builds with the 0.25 ryf configuration.

---

Using this existing configuration https://github.com/ACCESS-NRI/access-om3-configs/commit/58427153cd8250cdc35af7c469394b50948789c7 and the 0.3.0 build, `ocean.stats` shows,

```
Step,       Day,  Truncs,      Energy/Mass,      Maximum CFL,  Mean Sea Level,  Total Mass,  Mean Salin, Mean Temp, Frac Mass Err,   Salin Err,    Temp Err 
            [days]                 [m2 s-2]           [Nondim]       [m]             [kg]         [PSU]      [degC]       [Nondim]        [PSU]        [degC]
     0,  693135.000,     0, En 6.3048230599477494E-22, CFL  0.00000, SL  6.8348E-10, M 1.36283E+21, S 34.7205, T  3.6237, Me  0.00E+00, Se  0.00E+00, Te  0.00E+00
    63,  693136.000,     0, En 1.1422421593679974E-03, CFL  0.12730, SL  6.8277E-10, M 1.36283E+21, S 34.7206, T  3.6237, Me -3.81E-19, Se -2.83E-17, Te  8.84E-19
   127,  693137.000,     0, En 1.1320364471729537E-03, CFL  0.11451, SL  6.8348E-10, M 1.36283E+21, S 34.7206, T  3.6236, Me  1.59E-19, Se  1.36E-17, Te -1.75E-18
   191,  693138.000,     0, En 1.1510039944012112E-03, CFL  0.09915, SL  6.8065E-10, M 1.36283E+21, S 34.7206, T  3.6236, Me  7.92E-20, Se  5.60E-19, Te  2.54E-18
   255,  693139.000,     0, En 1.1290348792421638E-03, CFL  0.09146, SL  6.8277E-10, M 1.36283E+21, S 34.7206, T  3.6237, Me  4.93E-19, Se  1.61E-17, Te -2.30E-18
[...]

PR #726, opened 2025-08-19
Restart repro revisited

dougiesquire:
We have been compiling MOM in asymmetric memory mode in ACCESS-OM3 due to issues with restart reproducibility in symmetric memory mode. We never completely got to the root cause of the issue but it appeared to be related to the specific set of parameter values being used in the 100km configs (see [here](https://github.com/ACCESS-NRI/access-om3-configs/issues/447) and [here](https://github.com/mom-ocean/MOM6/issues/1660)). Our 25km configuration does not have the issue.

A lot has changed in MOM since we last tested the issue and my local testing suggests that the issue no longer exists. This PR is to test/document.

### DIFF
--- a/MOM_input
+++ b/MOM_input
@@ -141,6 +141,12 @@ DTFREEZE_DP = -7.75E-08         !   [degC Pa-1] default = 0.0
                                 ! When TFREEZE_FORM=LINEAR, this is the derivative of the freezing potential
                                 ! temperature with pressure.
 
+! === module MOM_restart ===
+RESTART_UNSIGNED_ZEROS = True   !   [Boolean] default = False
+                                ! If true, convert any negative zeros that would be written to the restart file
+                                ! into ordinary unsigned zeros.  This does not change answers, but it can be
+                                ! helpful in comparing restart files after grid rotation, for example.
+
 ! === module MOM_tracer_flow_control ===
 USE_IDEAL_AGE_TRACER = True     !   [Boolean] default = False
                                 ! If true, use the ideal_age_example tracer package.
--- a/MOM_input
+++ b/MOM_input
@@ -141,12 +141,6 @@ DTFREEZE_DP = -7.75E-08         !   [degC Pa-1] default = 0.0
                                 ! When TFREEZE_FORM=LINEAR, this is the derivative of the freezing potential
                                 ! temperature with pressure.
 
-! === module MOM_restart ===
-RESTART_UNSIGNED_ZEROS = True   !   [Boolean] default = False
-                                ! If true, convert any negative zeros that would be written to the restart file
-                                ! into ordinary unsigned zeros.  This does not change answers, but it can be
-                                ! helpful in comparing restart files after grid rotation, for example.
-
 ! === module MOM_tracer_flow_control ===
 USE_IDEAL_AGE_TRACER = True     !   [Boolean] default = False
                                 ! If true, use the ideal_age_example tracer package.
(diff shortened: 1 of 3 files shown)


dougiesquire:
!test repro

dougiesquire:
!test repro

dougiesquire:
```diff
$ git diff --no-index /scratch/tm70/repro-ci/experiments/access-om3-configs/f9ac6dd1132146c4cb0226509121d85d36355c1d/restart-1d-1-checksum.json /scratch/tm70/repro-ci/experiments/access-om3-configs/f9ac6dd1132146c4cb0226509121d85d36355c1d/restart-2d-0-checksum.json
diff --git a/scratch/tm70/repro-ci/experiments/access-om3-configs/f9ac6dd1132146c4cb0226509121d85d36355c1d/restart-1d-1-checksum.json b/scratch/tm70/repro-ci/experiments/access-om3-configs/f9ac6dd1132146c4cb0226509121d85d36355c1d/restart-2d-0-checksum.json
index 3b7d4ac..fbb5890 100644
--- a/scratch/tm70/repro-ci/experiments/access-om3-configs/f9ac6dd1132146c4cb0226509121d85d36355c1d/restart-1d-1-checksum.json
+++ b/scratch/tm70/repro-ci/experiments/access-om3-configs/f9ac6dd1132146c4cb0226509121d85d36355c1d/restart-2d-0-checksum.json
@@ -77,7 +77,7 @@
       "5A6E014229DC2D9"
     ],
     "v": [
-      "49C837CF6755FC84"
+      "C9C837CF6755FC84"
     ],
     "v2": [
       "4EB49AF59217D3C2"
```

This is the ["+/- 0.0 issue"](https://github.com/ACCESS-NRI/model-config-tests/issues/155). So **we now have restart reproducibility** :).

I thought the "+/- 0.0 issue" would be fixed by a newly introduced MOM6 parameter, `RESTART_UNSIGNED_ZEROS`, that converts any negative zeros that would be written to the restart file into ordinary unsigned zeros. However unfortunately this doesn't play nicely with `RESTART_CHECKSUMS_REQUIRED` as I'll demonstrate now.

dougiesquire:
From `/scratch/tm70/repro-ci/experiments/access-om3-configs/45399c19715f4ea4de2d27894d052729dbc93bd3/control/exp_1d_runtime/access-om3.err`:

```
FATAL from PE     0: MOM_restart(restore_state): Checksum of input field u BA9B16815B925DED does not match value 3A9B16815B925DED stored in ./access-om3.mom6.r.1900-01-02-00000.nc
```

This is because with `RESTART_UNSIGNED_ZEROS=True`, negative zeros are set to positive in the [`MOM_write_field` routine](https://github.com/ACCESS-NRI/MOM6/blob/8deb98524f2326c03a05db6516fd8b5b8acf01b6/src/framework/MOM_restart.F90#L1761-L1783), which is called _after_ the [restart checksums are calculated](https://github.com/ACCESS-NRI/MOM6/blob/8deb98524f2326c03a05db6516fd8b5b8acf01b6/src/framework/MOM_restart.F90#L1715-L1751). So the checksum in the restart file contains negative zeros and the validation checksum calculated when reading the restart does not, causing MOM6 to abort when `RESTART_CHECKSUMS_REQUIRED=True`.

I think I'll need to follow up with MOM6 devs about why this was implemented this way.

dougiesquire:
Closing as this has served its purpose

PR #753, opened 2025-09-11
Update mkdocs.yml for stub management

chrisb13:
closes https://github.com/ACCESS-NRI/access-om3-configs/issues/752

Issue #752, opened 2025-09-11
Update config stub management from OSIT ACCESS workshop meeting


chrisb13:
As discussed at workshop @chrisb13 @anton-seaice @minghangli-uni @atteggiani @aekiss @dougiesquire 

![Image](https://github.com/user-attachments/assets/99c55d00-1201-4908-a2d0-5b0cd0fbed1b) 

See design/discussion doc [here](https://anu365-my.sharepoint.com/:w:/g/personal/u1164007_anu_edu_au/EbSml9JXorVBvPjOt05Hmb8BASK5dtGiiDjTUYDgVwEKjQ).

chrisb13:
See discussion [here](https://forum.access-hive.org.au/t/cosima-twg-meeting-minutes-2025/4067/21?u=cbull).

In short
> We are going to move the all the documentation stubs back into main branch. 

[I've opened this PR](https://github.com/ACCESS-NRI/access-om3-configs/pull/796) so that we can work towards closing this?

I guess we need to move everything off stubs?

There's also this content: https://github.com/ACCESS-NRI/access-om3-configs/pull/573

chrisb13:
@anton-seaice as agreed at last week's TWG. I think it would be helpful to document where we got to last week and the plan going forwards. 

Can you please write some notes here on that?

It would also be good to get the ball rolling on what we need for versioning of the docs to be integrated / displayable ([this is likely relevant](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/)).

atteggiani:
It seems to me we are going in circle as well. 
I repeat it here if it was not clear during the many meetings and written posts.

The problem is neither which framework we use to build the website (mkdocs vs readthedocs) nor what framework we use to host the website (GitHub pages vs readthedocs) nor how we build and host it (custom GitHub action vs default `gh-pages` method).

The versioning problem comes from the actual structure of the versions (multiple independent versions within the same repository and therefore website). In addition to this the stubs infrastructure make everything more complicated.

anton-seaice:
Thanks Davide - we are going to stop using the stubs , and version the docs with CalVer versions (probably triggered by pushing a git tag). So all the docs will live in the main branch and the website gets built from that.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

anton-seaice:
[...]

PR #761, opened 2025-09-16
hotfix 25km release with documentation stub

anton-seaice:
This is a documentation only change to allow the documentation stub to be built from a tag in the config docs website. (A precursor to #753)

contributes to #752


anton-seaice:
See example of build with this stub : https://access-om3-configs.access-hive.org.au/pr-preview-762/configurations/MC_25km_jra_ryf/

anton-seaice:
I will manually make a tag after this PR is merged:

`release-MC_25km_jra_ryf-1.0-beta-1`

If we increment the version then we need a new DOI and to update the version numbers in the branch etc

dougiesquire:
> If we increment the version then we need a new DOI and to update the version numbers in the branch etc

Here you're referring to if we went to `1.1-beta`?

dougiesquire (approved):
LGTM thanks @anton-seaice

chrisb13:
Thanks @anton-seaice. Not sure if you saw it? But here's [my post on Zulip from Monday](https://access-nri.zulipchat.com/#narrow/channel/470332-ocean-seaice/topic/OM3.20config.20and.20wombat.20docs/near/539461107) about this:

>Chris Bull: @Anton had a chat to @Davide Marchegiani today as discussed. We noted:
> - could use an annotated (either by CI or by hand) tag to indicate which commit hash of the main docs was used on a branch when releasing a stub. Could alternatively use a git notes but these can be changed so it's probably a less useful idea.
> - davide suggested having a doc version in the tag (see next comment) and I think I agree -- this means we can do doc stub updates without changing the config version number;
> - need to be weary of having links to the main website inside stubs, i.e., when the links break if the website is updated there isn't an easy way for the user to render that version of the website. Perhaps we just have a policy of not having links from the main website to the stubs and vice versa.

And Re: need to be weary of having links to the main website inside stubs
> An example is [here](https://access-om3-configs.access-hive.org.au/configurations/dev-MC_25km_jra_ryf/). See how the following text has a broken link

I mentioned this last point to @minghangli-uni (in person), so I think we'd need to do a careful check to make sure there arn't any others. 

I'm personally in favour of a policy of not having links from the main website to the stubs and vice versa.

chrisb13:
> > > If we increment the version then we need a new DOI and to update the version numbers in the branch etc
> > 
> > 
> > Here you're referring to if we went to `1.1-beta`?
> 
> Yes

I'm a bit lost on this. So is the proposal to have:
 1. `release-MC_25km_jra_ryf-1.0-beta-1`
 1. `release-MC_25km_jra_ryf-1.1-beta`
 1. something else?

I think the first option is more useful (see comment above) as it versions the stub on it's own.

anton-seaice:
Yes number 1 is preferred

chrisb13:
ok

> See example of build with this stub : https://access-om3-configs.access-hive.org.au/pr-preview-762/configurations/MC_25km_jra_ryf/

So is the content exactly the same here?

anton-seaice:
I [added](https://github.com/ACCESS-NRI/access-om3-configs/pull/762/files) the target branch (752-release-MC_25km_jra_ryf) to that specific build (PR #762) of the preview website. So this should be equivalent to building the website with a tag made from this PR (as will happen soon).

anton-seaice:
OK ill merge this so I can progress it - comment in #752 if there is stuff that doesn't work

chrisb13:
> OK ill merge this so I can progress it 

I would have preferred to fix the known bugs, [example](https://github.com/ACCESS-NRI/access-om3-configs/pull/761#issuecomment-3301022879):

> An example is [here](https://access-om3-configs.access-hive.org.au/configurations/dev-MC_25km_jra_ryf/). See how the following text has a broken link

Can such fixes still be made?

anton-seaice:
Oh sorry - I see now. I suggest just fixing in the dev branch and otherwise ignoring it

anton-seaice:
Just hold off on 1 until after TWG on Wednesday. (the change would need to go into both dev and release branches, the order doesn't matter a lot in terms of which branch its done in first).

For 2. No it hasn't been implemented, and we decided not to do anything about it for now. e.g. I thought the opposite had been agreed upon. 

Anyway, both problems go away if we just move the stub content to the main branch. So its on the agenda for TWG again.

anton-seaice:
The issue with links between studs and main is new.

Its not that much has changed, I just think the case can be made more strongly since we decided automatic versioning of the main docs into config releases is too hard.

PR #807, opened 2025-10-02
Updates to the include-stubs plugin

atteggiani:
Updates the naming of the `include-stubs` plugin, which was renamed from `include-configuration-stubs`.


chrisb13 (approved):
Thaniks for addressing this so quickly!

PR #844, opened 2025-10-21
Update Edit-Config-Docs.md

Whyborn:
Remove section on not having write privileges due not being part of ACCESS-NRI due to github management changes

**1. Summary**:

Tiny adjustment in the docs- since we changed the way we manage Github permissions, the statement

**No write access**: (e.g. you are not part of the `ACCESS-NRI` Github organisation)

seems no longer quite accurate. Came across this when using the OM3 docs as a starting point for the AM3 docs. Suggested change is to remove the bracketed part, maybe add a bit on asking for privileges?

**9. Merge Strategy**

- [x] Merge commit


chrisb13:
> Yea, I don't have a merge button of any kind

The irony of this is good, no?

Thanks @Whyborn and @anton-seaice. Note the change @ccarouge given we're about to pick it up in esm1.6.

PR #849, opened 2025-10-23
Adding license https://github.com/ACCESS-NRI/access-om3-configs/issue…

### LINKED ISSUE #848: Should we have a licence in this repo?
@aidanheerdegen anyone: should we have a licence in this repo?

I note this doesn't have one either: https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments
Was that intended @ccarouge?

chrisb13:
Closes https://github.com/ACCESS-NRI/access-om3-configs/issues/848#issuecomment-3435126327


chrisb13:
> do we need different licenses for the github workflows compared to the docs

I had a bad feeling you might say this!

I don't mind but please update _all_ the following ones in line with what folks decide:
1. https://github.com/ACCESS-NRI/access-om3-configs/pull/849
1. https://github.com/ACCESS-Community-Hub/access-om3-experiments
1. https://github.com/access-nri/access-cm3-config-docs

>I note this doesn't have one either: https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments
Was that intended @ccarouge?

Note I didn't do anything about the above as we still haven't heard back from @ccarouge ?

Finally, maybe this discussion is best placed in [the issue](https://github.com/ACCESS-NRI/access-om3-configs/issues/848#issuecomment-3435126327)?

anton-seaice:
> I don't mind but please update _all_ the following ones in line with what folks decide:

@chrisb13 - this is not how reviews works - you need to follow through and finish work you start

chrisb13:
> > I don't mind but please update _all_ the following ones in line with what folks decide:
> 
> @chrisb13 - this is not how reviews works - you need to follow through and finish work you start

Well, in general yes but I think it's a bit convoluted in this case because those other repos don't have review processes and [I took your feedback](https://github.com/ACCESS-NRI/access-om3-configs/issues/848#issuecomment-3435004582) which following protocol was in an issue, as the way forward? 

In other words, when there's a change of course following consultation, who's responsibility is it to rectify it? (I don't know the answer to this...)

Given it's largely a docs issue, it's probably not our call to decide on the licence? Maybe [@ACCESS-NRI/hivedocsteam](https://github.com/orgs/ACCESS-NRI/teams/hivedocsteam)?

anton-seaice:
@chrisb13 - could you finish this or reassign please

chrisb13:
Followed up [here](https://access-nri.zulipchat.com/#narrow/dm/784078,784243,784244,784273,784274,784286-group/near/593667656).

chrisb13:
[After further discussion](https://access-nri.zulipchat.com/#narrow/dm/784078,784243,784244,784273,784274,784286-group/near/594189678), we've landed on Apache 2, which I've now added [0a4cd00](https://github.com/ACCESS-NRI/access-om3-configs/pull/849/commits/0a4cd00c737943f4fff27f65a4b09699a48261ea). 

@anton-seaice okay to proceed? Would suggest we re-base. I think that will create a merge commit (which will then have a ccby commit in it) if we use GitHub. Perhaps I should do a re-base offline, then a force push, before we do a squash merge? Unless there's a better way? Another option would just be to start again off the current main...

anton-seaice:
Looks good. There's no conflicts, so just hitting `Squash and merge` will be fine

anton-seaice (approved):

chrisb13:
> There's no conflicts, so just hitting `Squash and merge` will be fine

Ah good point, much simpler, let's do that.

Issue #332, opened 2024-04-23
Release Procedure for OM3


anton-seaice:
We should write a release procedure for ACCESS-OM3, encapsulating the related spack / build and configuration steps to produce a coherant / complete release.

I think this would cover:
- Tagging and versioning the build (i.e. this repo).
- Building through spack, and creating / moving to the final location if needed
- Copying the inputs to a release folder (i.e. with a version number)
- Updating all the configuration repositories for release (including updating `config.yaml` to turn on runlog, collate, metadata, reproducibility, post-processing etc and any other updates needed for versioning. Possibly also change run frequencies and turn off debug logging.)
- Testing

@micaeljtoliveira has suggested we could write this in the form of an issue template (I think it would in main of this repository).

anton-seaice:
Leaving notes for myself:

the process for updating CITATION.cff is (currently) an all manual step before release, but can be automated in the future

anton-seaice:
I think between 

https://access-om3-configs.access-hive.org.au/infrastructure/Building/#new-software-deployments

and 

https://forum.access-hive.org.au/t/model-release-checklist-template/4371

We have this covered these days

chrisb13:
> We have this covered these days

Agreed, best not to duplicate.

PR #877, opened 2025-11-04
Note about release checklist

anton-seaice:
Add release checklist to docs (related to #332)


anton-seaice:
Ready for another look @chrisb13

anton-seaice:
> Just out of interest, re: this [link](https://zenodo.org/communities/access-nri/records?q=conceptdoi%3A16294824&l=list&p=1&s=10&sort=bestmatch), what will it include going forwards? Any om3 release?

There's two DOIs, one is the concept DOI and the other is the one specific to each version. The concept DOI is the same (well probably) for all config releases, and the versioned one is unique to each release

This link searches for all versioned DOI's which use the concept DOI. (Just using the concept DOI goes to the latest release, which may come from any branch)

chrisb13:
@anton-seaice my impression from [the recent zulip chat](https://access-nri.zulipchat.com/#narrow/channel/470332-ocean-seaice/topic/DOI.20.2F.20zenodo.20for.20config.20releases/with/554611000) is that this needed to be re-visited or are we ok?

PR #883, opened 2025-11-13
Add commit signing to cherry-pick

anton-seaice:
@CodeGat 

Is it as simple as this to add commit signing to the cherry-pick workflow?

How would i test this?


anton-seaice:
Living dangerously here ! 😬

PR #923, opened 2025-11-27
Updated RTD setup files

atteggiani:
- [x] Updated github pages deployment workflow
- [x] Updated RTD config
- [x] Added link-checking
- [x] Updated ci.yml to prevent it from running when only files related to documentation are changed

My admin permissions can be revoked now (please leave write permissions).


atteggiani:
> There appear to be three or four unrelated changes here? Can there please be one PR for each standalone/independent change with a link to which issue they are related to. This will make the scope of the changes clearer currently and in the future, and easier to review.

The changes are all related to the deployments.
Some are related to the GitHub pages deployment (needed because some changes are important to work together with RTD) and some to the RTD deployment.

> I don't think moving the `.readthedocs.yaml` file into a documentation folder is a good idea. This means we need a `documentation` folder in all the config branches - which do not contain documentation

This is a good point. I'll move the `.readthedocs.yaml` back to the root repo.

atteggiani:
> The link checker, updating the RTD configuration and updating the pages deployment are three unrelated pieces of work.

They are all related to how you build the site, especially to the `READTHEDOCS_CANONICAL_URL` env variable. But I can try and separate them into 3 different PRs if you prefer.
I'm noticing now this repo doesn't have the `resolve_absolute_urls` plugin, do you know why it was removed?

atteggiani:
> > I'm noticing now this repo doesn't have the `resolve_absolute_urls` plugin, do you know why it was removed?
> 
> I can't see it was ever in the history of https://github.com/ACCESS-NRI/access-om3-configs/blob/main/documentation/requirements.txt
> 
> Is it a python plugin which would go in that file ?

Yes, and included in the `mkdocs.yml`. I'll include it in a PR

Issue #901, opened 2025-11-20
enable link checking for the access om3 config docs


chrisb13:
Following up on discussion with @atteggiani [here](https://access-nri.zulipchat.com/#narrow/channel/470303-nri-team/topic/Hive.20Docs.20pages.20structure.20change/near/555500543) this issue (subsequent PR) is to enable link checking for the access om3 config docs.

Thanks to @aekiss for prompting us to get this working again.

PR #927, opened 2025-11-28
Link checker

atteggiani:
> [!IMPORTANT]
> ~This PR HEAD branch was based off #925's head branch. Please only merge this PR once #925 is merged and then this head branch has been rebased to `main`~.
> The branch has been rebased and this PR can now be reviewed/merged.

- [x] Added link checker
- [x] Updated mkdocs hook to also change canonical url

### DIFF
--- a/documentation/docs/pages/inputs/Forcing-data-models.md
+++ b/documentation/docs/pages/inputs/Forcing-data-models.md
@@ -7,15 +7,15 @@ Forcing is provided via [CDEPS](https://github.com/ESCOMP/CDEPS) data models [do
 
 ## Coupling
 
-- DATM and DROF are coupled to the other components via the mediator - see the [coupling architecture here](../infrastructure/Architecture.md).
+- DATM and DROF are coupled to the other components via the mediator - see the [coupling architecture here](/infrastructure/Architecture.md).
 - The coupled fields and remapping methods used are recorded in the mediator log output file and can be found with `grep '^ mapping' archive/output000/log/med.log`; see [here](https://escomp.github.io/CMEPS/versions/master/html/esmflds.html) for how to decode this.
-- See [the Configurations Overview page](../configurations/Overview.md#coupling) for details on how the coupling is determined.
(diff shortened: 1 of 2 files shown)


anton-seaice:
This contributes to #901

atteggiani:
> Can we tidy up the logging in `check-links` so that only shows errors?

There is the [Summary](https://github.com/ACCESS-NRI/access-om3-configs/actions/runs/19750849691?pr=927#summary-56593523236) for that

atteggiani:
> I think we will need some repository specific exceptions? What are your thoughts ?

Yes, some links might need specific exceptions and these seem good cases to me. 

The first (https://forum.access-hive.org.au/t/model-release-checklist-template/4371) goes to a private forum post, so `lychee` cannot see it. 
As lychee, users that don't have permissions (I believe most of the users) would also not be able to see it. Do we really want to link to that post on a public website?
If the answer is yes, we can set `lychee` config to ignore this specific link.

For the second (https://mermaid.live/edit#pako:eNptkDFrwzAQhf-KuNE4awcPXZIUOpgO7dQqhEOSY4HsC4pEJeL891wSW4USTe8-3jtO7wyKtIEGOke_qkcfxNdGjoIf_VRV-9G-VNXuASyD9ft6y0QodCo6DOYk3izZfcCY6iLznEi3FUZbDOT_hZDSEkLK9xBLSoXOsqwSq9WrmIpjoqc8L9zOfDluSk95nhLUMBg_oNVcw_nmkhB6MxgJDUttOowuSJDjha0YA33mUUETfDQ1eIqHHpoO3YmneNT8vY3Fg8eh0COO30R_MzfChbSP4u_9X65dmIJi) the problem I believe is that this link is a live editor that most-likely gets its token and url slug from JS, so the static page itself at that link might not exist (so `lychee` cannot find it). 
This cannot be fixed and I would not set a specific ignore only for this link. What I would do instead, for cases like this, is simply ignore any `https://mermaid.live/edit#...` link.


What do you think?

PR #928, opened 2025-11-28
Added resolve-absolute-urls plugin

atteggiani:
- [x] Added [mkdocs_resolve_absolute_urls](https://github.com/ACCESS-NRI/mkdocs_resolve_absolute_urls_plugin) plugin, to allow absolute urls to be correctly resolved.


atteggiani:
> Appears to be [failing](https://github.com/ACCESS-NRI/access-om3-configs/actions/runs/19750926853/job/56593714319?pr=928) ? @atteggiani

Working now. Had to be rebased to `main`.

chrisb13:
> This change I think is needed so that absolute links can be used and generate the correct URLs in PR previews.

That would be nice to have I think.

PR #944, opened 2025-12-01
Release proc

anton-seaice:
Possible notes when conducting releases


anton-seaice:
I am not sure this is the right place for it but well give it a think

chrisb13:
> I am not sure this is the right place for it but well give it a think

Yeah I wondered that as well. I think it depends, if it's OSIT specific then I think it's an okay home. If it's more broadly relevant than perhaps Aidan et al.'s checklist?

PR #982, opened 2025-12-10
Remove -nomask for atmosphere

anton-seaice:
**1. Summary**:

What has changed?

Why was this done?

**2. Issues Addressed:**

-

**3. Dependencies (e.g. on payu, model or om3-scripts)**

This change requires changes to (note required version where true):
- [ ] payu:
- [ ] access-om3:
- [ ] om3-scripts:

**4. Ad-hoc Testing**

What ad-hoc testing was done? How are you convinced this change is correct (plots are good)?

**5. CI Testing**

- [ ] `!test repro` has been run

**6. Reproducibility**

- [ ] Yes
- [ ] No - `!test repro commit` has been run.  

**7. Documentation**

- [ ] Yes
- [ ] N/A

- [ ] Yes: 
- [ ] N/A

**8. Formatting**

- [ ] Yes
- [ ] N/A

**9. Merge Strategy**

- [ ] Merge commit
- [ ] Rebase and merge
- [ ] Squash

### DIFF
--- a/testing/checksum/historical-6hr-checksum.json
+++ b/testing/checksum/historical-6hr-checksum.json
@@ -2,100 +2,100 @@
   "schema_version": "1-0-0",
   "output": {
     "CAu": [
-      "FD6EB6E16FE3A618"
+      "32703EF6B5437A3A"
     ],
     "CAv": [
-      "876CD3235EAABCC7"
+      "6E04714E13C043BF"
     ],
     "DTBT": [
-      "4055967BB1F3337D"
+      "4055967BB1EE96C0"
     ],
     "First_direction": [
       "0"
     ],
     "Kd_shear": [
-      "92681DF21667F176"
+      "1228D436721F66B2"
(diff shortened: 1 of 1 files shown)


anton-seaice:
!test repro

minghangli-uni:
so it appears cmeps can auto-detect if the mesh is the same or not

anton-seaice:
Yeah just a demonstration at this point

anton-seaice:
But will probably make this change soon

anton-seaice:
In the esmf docs it says:

> Higher-order patch recovery (Patch): Patch rendezvous method of taking the least squares fit of the surrounding surface patches. This is a higher order method that may produce interpolation weights that may be slightly less than 0 or slightly greater than 1.


So maybe that's enough to introduce this level of differences (old uses patch remapping from -nomask to mask mesh, which may be close to 1, where new will be exactly 1)

PR #1033, opened 2026-01-10
Testing model build 2026.01.000 on `dev-MC_100km_jra_ryf`

### LINKED ISSUE #1032: Testing model upstream component and dependencies update: 2026.01.000
Issue to discuss the testing of new model build 2026.01.000.

chrisb13:
Closes #1032. PR to test new om3 build 2026.01.000.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -19,7 +19,7 @@ modules:
     use:
         - /g/data/vk83/prerelease/modules
     load:
-        - access-om3/pr176-16
+        - access-om3/pr176-17
 
 queue: normal
 ncpus: 240
--- a/config.yaml
+++ b/config.yaml
@@ -19,7 +19,7 @@ modules:
     use:
         - /g/data/vk83/prerelease/modules
     load:
-        - access-om3/pr176-18
+        - access-om3/pr176-19
 
 queue: normal
 ncpus: 240
--- a/config.yaml
+++ b/config.yaml
@@ -19,7 +19,7 @@ modules:
     use:
         - /g/data/vk83/prerelease/modules
     load:
-        - access-om3/pr176-19
+        - access-om3/pr176-20
 
 queue: normal
 ncpus: 240
--- a/config.yaml
+++ b/config.yaml
@@ -19,7 +19,7 @@ modules:
     use:
         - /g/data/vk83/prerelease/modules
     load:
-        - access-om3/pr176-17
+        - access-om3/pr176-18
 
 queue: normal
 ncpus: 240
--- a/config.yaml
+++ b/config.yaml
@@ -1,5 +1,5 @@
 # Payu configuration
-
+ 
 # If submitting to a different project to your default, uncomment line below 
 # and change project code as appropriate; also set shortpath below
 # project: x77
--- a/config.yaml
+++ b/config.yaml
@@ -19,7 +19,7 @@ modules:
     use:
         - /g/data/vk83/prerelease/modules
     load:
-        - access-om3/pr176-2
+        - access-om3/pr176-6
 
 queue: normal
 ncpus: 240
--- a/config.yaml
+++ b/config.yaml
@@ -19,7 +19,7 @@ modules:
     use:
         - /g/data/vk83/prerelease/modules
     load:
-        - access-om3/pr176-12
+        - access-om3/pr176-13
 
 queue: normal
 ncpus: 240
--- a/config.yaml
+++ b/config.yaml
@@ -19,7 +19,7 @@ modules:
     use:
         - /g/data/vk83/prerelease/modules
     load:
-        - access-om3/pr176-6
+        - access-om3/pr176-12
 
 queue: normal
 ncpus: 240
--- a/config.yaml
+++ b/config.yaml
@@ -19,7 +19,7 @@ modules:
     use:
         - /g/data/vk83/prerelease/modules
     load:
-        - access-om3/pr176-13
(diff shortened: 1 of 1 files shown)


dougiesquire:
!test repro

anton-seaice:
> Can someone tell me why this failed? [Here's the error](https://github.com/ACCESS-NRI/access-om3-configs/pull/1033/checks?check_run_id=59974595965):
> 
> ```python
> AssertionError: Failed to find release artefact for model version at https://github.com/ACCESS-NRI/ACCESS-OM3/releases/tag/pr176-2
> assert 404 == 200
> ```
> 
> Rather reads like it can't find my [pre-release](https://github.com/ACCESS-NRI/ACCESS-OM3/pull/176#issuecomment-3731707830)?
> 

This just mean's it can't find the pre-release in the ACCESS-OM3 releases - which is expected obviously

anton-seaice:
> > Can someone tell me why this failed? [Here's the error](https://github.com/ACCESS-NRI/access-om3-configs/pull/1033/checks?check_run_id=59974595965):
> 
> You need to update the Payu manifests: `payu setup && payu sweep` (note that if we don't have `metadata:enable:false` in the `config.yaml` then doing this will add a commit)

I don't think this is correct, exe checksums are not being checked, (it's turned off)

From config.yaml:

```
manifest:
  reproduce:
    exe: false
```

dougiesquire:
> I don't think this is correct, exe checksums are not being checked, (it's turned off)
> 
> From config.yaml:
> 
> ```
> manifest:
>   reproduce:
>     exe: false
> ```

I'm trying to say that if `metadata:enable:false` is not set, then doing `payu setup` will add a commit.

chrisb13:
Thanks @anton-seaice and @dougiesquire. Sweet, confirmed (expected) that [openmpi going from](https://github.com/ACCESS-NRI/ACCESS-OM3/pull/176/changes) `4.1.7` to `5.0.8` [does not change answers](https://github.com/ACCESS-NRI/access-om3-configs/pull/1033#issuecomment-3735970756). Few questions:

> You need to update the Payu manifests: `payu setup && payu sweep`

So did you do this? (I think not.) Is it something in [here](https://github.com/ACCESS-NRI/access-om3-configs/tree/1032-model-buid-update-test/manifests)? Normally,  `payu setup && payu sweep` is something I do in a cli so I'm not sure how it applies here.

> This just mean's it can't find the pre-release in the ACCESS-OM3 releases - which is expected obviously

Ok, so effectively the `payu setup && payu sweep` bypasses this check somehow?

dougiesquire:
> > You need to update the Payu manifests: `payu setup && payu sweep`
> 
> So did you do this? (I think not.) Is it something in [here](https://github.com/ACCESS-NRI/access-om3-configs/tree/1032-model-buid-update-test/manifests)? Normally, `payu setup && payu sweep` is something I do in a cli so I'm not sure how it applies here.

Yeah, you need to do this at the command line. 

> > This just mean's it can't find the pre-release in the ACCESS-OM3 releases - which is expected obviously
> 
> Ok, so effectively the `payu setup && payu sweep` bypasses this check somehow?

Sorry, it was misleading for me to make my comment a reply to yours. Updating the manifest will not fix the `test_access_om3_manifest_exe_in_release_spack_location` error

dougiesquire:
> I'm confused, so what does `payu setup && payu sweep` achieve here? Are you suggesting I could go to a cli, checkout `1032-model-buid-update-test` do `payu setup && payu sweep` and then commit the change? (I would have guessed it wouldn't result in a change but would check that the config can find input files etc.)

It will mean that the manifests are correct. It's good to get in the habit of doing this. E.g. in many of our other configs we have:
```
manifest:
  reproduce:
    exe: True
    input: True
```
 so updating exe/inputs and failing to update the manifests would mean that the config won't run.
 
 I was trying to give some general helpful advice, but it would appear I've only added confusion...

> On the `test_access_om3_manifest_exe_in_release_spack_location` error, we just ignore that then because this is a pre-release?

Yup

dougiesquire:
> !test repro
> 
> Testing generic tracers

Just a note that this configuration does not actually have WOMBAT configured

dougiesquire:
> @dougiesquire suggested `manifest/exe` but I think that will fail here, is that a concern (or a feature)?

I don't think I did suggest this.

> So, I'm suggesting we could turn on runlog and manifest/inputs
> Maybe we don't want runlog in this context?

There's not much value in setting either of these in this PR if that's what you're asking?

dougiesquire:
My (intended) point was that it's probably good practice to keep the configuration manifests up-to-date with configuration changes

chrisb13:
Okay from [this test](https://github.com/ACCESS-NRI/access-om3-configs/pull/1033#issuecomment-3766449524).

It seems 

>❌ test_repro_historical
✅ test_repro_determinism
✅ test_repro_restart

I was expecting the first one to pass as well. 

The changes are here (I think!):
https://github.com/ACCESS-NRI/ACCESS-OM3/pull/176/changes

which means it's either `access3-share`, `spack-packages` or the combination of previous changes that were not answer changing (`access-generic-tracers` and/or `openmpi`).

Any guesses on what I should try first? Or I can just brute force it...

dougiesquire:
Almost certainly the `access3-share` update

chrisb13:
!test repro

chrisb13:
!test repro

chrisb13:
!test repro

Issue #1028, opened 2026-01-05
Upstream model component update: January 2026.


chrisb13:
This issue is intended as a place to discuss which upstream model versions we should update to for the new `ACCESS-OM3` build.

From [this](https://github.com/ACCESS-NRI/ACCESS-OM3/blob/main/spack.yaml), here are the current versions

```yaml
  specs:
    - access-om3@git.2025.08.001
  packages:
    # Main Dependencies
    access3:
      require:
        - '@2025.08.000'
        - configurations=MOM6,MOM6-CICE6,MOM6-CICE6-WW3
    access-cice:
      require:
        - '@CICE6.6.1-0'
    access-mom6:
      require:
        - '@2025.07.000'
    access-ww3:
      require:
        - '@2025.08.000'
    access3-share:
      require:
        - '@2025.08.000'
    access-generic-tracers:
      require:
        - '@2025.08.000'
    access-mocsy:
      require:
        - '@2025.07.002'
    # Other Dependencies
    esmf:
      require:
        - '@8.7.0'
    parallelio:
      require:
        - '@2.6.2'
    netcdf-c:
      require:
        - '@4.9.2'
    netcdf-fortran:
      require:
        - '@4.6.1'
    fms:
      require:
        - '@2025.03'
    openmpi:
      require:
        - '@4.1.7'
    fortranxml:
      require:
        - '@4.1.2'
    gcc-runtime:
      require:
        - '%gcc'
```

Related: 
 - [#ocean-seaice > Upstream model component updates](https://access-nri.zulipchat.com/#narrow/channel/470332-ocean-seaice/topic/Upstream.20model.20component.20updates/with/560491733)
 - https://access-om3-configs.access-hive.org.au/infrastructure/Building/#new-software-deployments
 - https://forum.access-hive.org.au/t/osit-team-roles-roster-v3-draft/4590
 - https://github.com/ACCESS-NRI/access-om3-configs/pull/1029
 - https://github.com/ACCESS-NRI/access-spack-packages/tree/main/packages

chrisb13:
We have the following components:
 - access3
 - access-cice
 - access-mom6
 - access-ww3
 - access3-share
 - access-generic-tracers
 - access-mocsy
 - esmf
 - parallelio
 - netcdf-c
 - netcdf-fortran
 - fms
 - openmpi
 - fortranxml
 - gcc-runtime

[...]

PR #1039, opened 2026-01-12
Update upstream-model-component-update-.md

chrisb13:
As suggested by @dougiesquire https://github.com/ACCESS-NRI/access-om3-configs/issues/1028#issuecomment-3736569758


dougiesquire:
@chrisb13, the row you added had `access-generic-tracers` instead of `access-spack-packages`. I've fixed up and moved into a new "Infrastructure" table

PR #1082, opened 2026-01-22
Add bash function for quick Payu cloning

chrisb13:
**1. Summary**:

Very small PR to add a bash function I wrote. Will create a conflict with @minghangli-uni's PR that adds to the same page. Perhaps we wait for it to be merged?


chrisb13:
This has been picked up here:
 - https://github.com/ACCESS-NRI/access-om3-configs/pull/1084
 - https://github.com/ACCESS-NRI/access-om3-configs/issues/1083

PR #1087, opened 2026-01-22
Updates to generic regional config to work with automated regional mom6 workflow

helenmacdonald:
@ashjbarnes have updated the generic stencil to:
1.  remove unnecessary files (that are needed for wombat or CICE but not mom-only)
2. change config parameters to align with the example regional was fig being used


helenmacdonald:
All good @ashjbarnes - we think the issues [was something else](https://github.com/ACCESS-NRI/access-om3-configs/issues/1088) and found a way to work around it. I want to do a couple of changes and then you are good to go.

PR #1099, opened 2026-01-29
Cherry pick from pr1092 into dev mc 100km jra iaf+wombatlite

anton-seaice:
Use better data meshes and turn off Initial conditions


dougiesquire (approved):
Ta

PR #1243, opened 2026-03-25
Update to ACCESS-OM3 2026.03.000, per #1028

anton-seaice:
Cherry pick new build (from #1225) 

> FATAL from PE     1: ocean_model_restart was called with unused buoyancy fluxes.  For conservation, the ocean restart files can only be created after the buoyancy forcing is applied.
> 
> FATAL from PE     2: ocean_model_restart was called with unused buoyancy fluxes.  For conservation, the ocean restart files can only be created after the buoyancy forcing is applied.
> 
> WARNING from PE     0: End of MOM_main reached with inconsistent dynamics and advective times.  Additional restart fields that have not been coded yet would be required for reproducibility.

?

### DIFF
--- a/testing/checksum/historical-6hr-checksum.json
+++ b/testing/checksum/historical-6hr-checksum.json
@@ -2,91 +2,91 @@
   "schema_version": "1-0-0",
   "output": {
     "Temp": [
-      "927714E14CC9C12D"
+      "924D5F90B5E0352E"
     ],
     "Salt": [
-      "4708042D7E2526F5"
+      "47080B6793AD5FD2"
     ],
     "h": [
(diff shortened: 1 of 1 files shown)


anton-seaice:
Does this mean anything to you @dougiesquire @helenmacdonald ?

anton-seaice:
Thanks @dougiesquire - another one :)

dougiesquire:
> For the OBCs, I think it makes sense to accept the bug fixes i.e. change to False - we don't have a run anyway that we want to preserve bitwise reproducibility, right?

Yup that's right. Thanks @claireyung!

PR #1280, opened 2026-04-17
Added information on finding executables using the 'which' command.

chrisb13:
Closes https://github.com/ACCESS-NRI/access-om3-configs/issues/1259


anton-seaice:
I think you want the `whereis` command rather than `which` but it's fairly moot.

I don't think this is specific enough to OM3 to go in our docs - its just general bash commands

PR #1288, opened 2026-04-22
update configs

aekiss:
**1. Summary**:

What has changed?

- ~increase collation memory to 120GB - https://github.com/ACCESS-NRI/access-om3-configs/issues/1013~
- set MAX_DELTA_SRESTORE = 1.0 - https://github.com/ACCESS-NRI/access-om3-configs/issues/1180
- set ADJUST_NET_SRESTORE_BY_SCALING = True - https://github.com/ACCESS-NRI/access-om3-configs/issues/260
- set NONBLOCKING_UPDATES = True for performance boost - https://github.com/ACCESS-NRI/access-om3-configs/issues/1245

Why was this done?

see issues

**2. Issues Addressed:**

see above

**3. Dependencies (e.g. on payu, model or om3-scripts)**

This change requires changes to (note required version where true):
- [ ] payu:
- [ ] access-om3:
- [ ] om3-scripts:

**4. Ad-hoc Testing**

What ad-hoc testing was done? How are you convinced this change is correct (plots are good)?

- 1-day test run
- Same parameters used for 3rd IAF test [`MC_25km_jra_iaf+wombatlite-test3v2-00532b88`](https://github.com/ACCESS-NRI/access-om3-configs/issues/1216#issuecomment-4210562635)

**5. CI Testing**

- [x] `!test repro` or `!test repro commit ` has been run

**6. Reproducibility**

Is this reproducible with the previous commit? (If not, why not?)
- [ ] Yes
- [x] No, not expected to be

**7. Documentation**

The docs folder has been updated with output from running the model?
- [ ] Yes
- [x] N/A

A PR has been created for updating the documentation?
- [ ] Yes: 
- [x] N/A

**8. Formatting**

Changes to MOM_input have been copied from model output in docs/MOM_parameter_docs.short?
- [x] Yes
- [ ] N/A

**9. Merge Strategy**

- [x] Merge commit
- [ ] Rebase and merge
- [ ] Squash  <--- might make it easier to cherry-pick to all configs?

Issue #1245, opened 2026-03-25
OM3 configs optimisation mega-issue


manodeep:
This issue will track the sub-issues for optimising individual configs, and note any insights that may (should?) apply generally across all configs.

## Individual optimisation issues:
- 25 km IAF config
- 100 km IAF config

## Performance insights
- Setting `NONBLOCKING_UPDATES =True` in `MOM_Input` results in a 10% performance boost for 25k IAF (25/03/2026)

manodeep:
@marshallward confirmed that the non-blocking updates feature is fully implemented and used in production runs

marshallward:
@manodeep Let me confirm this to be certain, I think I might have spoken too quickly when I was asked about this.

manodeep:
> [@manodeep](https://github.com/manodeep) Let me confirm this to be certain, I think I might have spoken too quickly when I was asked about this.

That will be great - thank you very much @@marshallward!

marshallward:
@manodeep I talked to a few people who confirmed that nonblocking updates are used in production.

manodeep:
Fantastic - thanks for confirming Marshall 🙂

dougiesquire:
We are finding that setting `NONBLOCKING_UPDATES = True` breaks restart-reproducibility (1x2day runs != 2x1day run) in our configs (see [issue here](https://github.com/ACCESS-NRI/access-om3-configs/issues/1312) and [test here](https://github.com/ACCESS-NRI/access-om3-configs/pull/1314)). So I think we will have to disable this for the time being.

manodeep:
Huh! So there is likely a difference between our setup (including our tests) and what GFDL is using. 

Is the repro failure indicating that something in the restart output routines is not quite right?

manodeep:
Now that I think a bit more - shouldn't that repro test failure have occurred *before* merging the PR, as in with a `!test repro` in the PR with only `NONBLOCKING_UPDATES=TRUE`?

anton-seaice:
> Now that I think a bit more - shouldn't that repro test failure have occurred _before_ merging the PR, as in with a `!test repro` in the PR with only `NONBLOCKING_UPDATES=TRUE`?

Yeah it did, we just missed it!

manodeep:
> > Now that I think a bit more - shouldn't that repro test failure have occurred _before_ merging the PR, as in with a `!test repro` in the PR with only `NONBLOCKING_UPDATES=TRUE`?
> 
[...]

Issue #1298, opened 2026-04-30
`dev-MC_4km_jra_ryf+regionalpanan` crashes in first few days


edoddridge:
### Issue description

The `dev-MC_4km_jra_ryf+regionalpanan` configuration crashes after three days of model time. 

There are some velocity truncations (see figure below), but the number returns to zero prior to the run crashing. Halving the tilmestep reduces the number of velocity truncations by ~10x, but the model still crashes in the first few days.

Using config as-is:
<img width="298" height="216" alt="Image" src="https://github.com/user-attachments/assets/e26e7ecd-52c4-4452-8b6a-39e1af3c9956" />

Overriding `DT` and `DT_THERM` to be 300s.
<img width="294" height="216" alt="Image" src="https://github.com/user-attachments/assets/44f5127c-a012-4523-a421-d67b210a72f4" />

### Issue checklist

 
- [ ] I have labelled this issue with the affected configurations (e.g. `MC_25km_jra_ryf`, [1] more options)
- [ ] I have added the issue information to the [ACCESS-OM3 Mega Project board!](https://github.com/orgs/ACCESS-NRI/projects/24):
After issue creation, under _Projects_ please set:
 - [ ] Status (blocked, Todo, In Progress, Done)
 - [ ] Priority (P0, P1, P2) 
 - [ ] Component

I don't appear to have the privileges to add a label. `dev-MC_4km_jra_ryf+regionalpanan` is the appropriate one to add.

dougiesquire:
Thanks for opening the issue @edoddridge. I'm curious, did you try checking out and running from the commit I suggested via email?

```
$ git checkout ac99c202a8e43eb2b8f40bc470dd9a6b3507a30f
```

anton-seaice:
> Halving the tilmestep reduces the number of velocity truncations by ~10x, but the model still crashes in the first few days.

Could you add the log message? i.e. do the truncations cause the crash, or something else ?

edoddridge:
As far as I can tell, the truncations **do not** cause the crash. I haven't been able to work out what does yet (but I'm not very experienced with MOM6/OM3 errors, so that's likely just user error).

I was partway through writing this and @anton-seaice popped by. Together we found 
```
FATAL from PE   298:  Could not find target coordinate   1033.00000000000      in get_polynomial_coordinate. This is caused by an inconsistent interpolant (perhaps not monotonically increasing):   1033.00000000000                          NaN   1037.02212973507
```
[...]

Issue #1251, opened 2026-03-27
ESMF error in dev-MC_25km_jra_iaf+wombatlite


aekiss:
dev-MC_25km_jra_iaf+wombatlite crashes on model date 1963-01-15

lots of `PET*.ESMF_LogFile` have errors such as this:

`/home/156/aek156/payu/MC_25km_jra_iaf+wombatlite/work/PET0007.ESMF_LogFile`
```
20260327 091958.467 ERROR            PET0007 ERROR: 135 nans found in So_u
20260327 091958.468 ERROR            PET0007 ERROR: 135 nans found in So_v
20260327 091958.468 ERROR            PET0007 med_methods_mod.F90:2475 ABORTING JOB, see PET file for details
20260327 091958.468 ERROR            PET0007 med_phases_prep_ice_mod.F90:154 Failure  - Passing error in return code
20260327 091958.468 ERROR            PET0007 MED:src/addon/NUOPC/src/NUOPC_ModelBase.F90:2104 Failure  - Passing error in return code
20260327 091958.468 ERROR            PET0007 ESM0001:src/addon/NUOPC/src/NUOPC_Driver.F90:3622 Failure  - Phase 'med_phases_prep_ice' Run for modelComp 1 did not return ESMF_SUCCESS
20260327 091958.468 ERROR            PET0007 ESM0001:src/addon/NUOPC/src/NUOPC_Driver.F90:3860 Failure  - Passing error in return code
20260327 091958.468 ERROR            PET0007 ESM0001:src/addon/NUOPC/src/NUOPC_Driver.F90:3536 Failure  - Passing error in return code
20260327 091958.468 ERROR            PET0007 ensemble:src/addon/NUOPC/src/NUOPC_Driver.F90:3622 Failure  - Phase 'RunPhase1' Run for modelComp 1 did not return ESMF_SUCCESS
20260327 091958.468 ERROR            PET0007 ensemble:src/addon/NUOPC/src/NUOPC_Driver.F90:3860 Failure  - Passing error in return code
20260327 091958.468 ERROR            PET0007 ensemble:src/addon/NUOPC/src/NUOPC_Driver.F90:3536 Failure  - Passing error in return code
20260327 091958.468 ERROR            PET0007 esmApp.F90:147 Failure  - Passing error in return code
20260327 091958.468 INFO             PET0007  Finalizing ESMF with endflag==ESMF_END_ABORT
```

[...]

PR #1309, opened 2026-05-01
Cherry pick from pr1302 into dev mc 4km jra ryf+regionalpanan

anton-seaice:
This sets the MPI argument `mpi_yield_when_idle` per #1245

This was tested as part of #1288 although setting the MOM parameter NONBLOCKING_UPDATES was found to be not reproducible over restarts (see #1312). So we only set the MPI argument here.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -62,10 +62,6 @@ platform:
     nodesize: 104
     nodemem: 512
 
-mpi:
-  flags:
-    - --mca pml ucx --mca coll ^ucc -x UCX_RNDV_THRESH=inf
-
 collate:
   restart: true
   mpi: true
@@ -77,7 +73,7 @@ collate:
 
 mpi:
   flags:
-  - --mca mpi_yield_when_idle true
+  - --mca pml ucx --mca coll ^ucc -x UCX_RNDV_THRESH=inf --mca mpi_yield_when_idle true
 
 userscripts:
     archive: /usr/bin/bash /g/data/vk83/apps/om3-scripts/payu_config/archive.sh
--- a/config.yaml
+++ b/config.yaml
@@ -75,6 +75,10 @@ collate:
   queue: normalsr
   exe: mppnccombine-fast
 
+mpi:
+  flags:
+  - --mca mpi_yield_when_idle true
+
 userscripts:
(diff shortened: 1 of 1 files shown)


anton-seaice:
!test repro commit

anton-seaice:
@helenmacdonald @dougiesquire - I think this change make sense for panant?


> adjustments to salt restoring to achieve zero net are made by scaling 
> values without moving the zero contour.

dougiesquire:
I don't think so. This configuration has `ADJUST_NET_SRESTORE_TO_ZERO = False`

helenmacdonald:
Thanks for checking @anton-seaice! I think we might have turned this one off for regional models because when boundary conditions are in play we are not expecting net zero salt flux.

anton-seaice:
!test repro commit

config.yaml
```diff
@@ -75,6 +75,10 @@ collate:
   queue: normalsr
   exe: mppnccombine-fast
 
+mpi:
+  flags:
+  - --mca mpi_yield_when_idle true
+
```
dougiesquire, line 77: There is already an `mpi:` section 13 lines above

dougiesquire (changes requested):

config.yaml
anton-seaice: Ok i've fixed this

anton-seaice:
This config crashed for me at `19000103       40200` , with the same crash as #1251 (different again to https://github.com/ACCESS-NRI/access-om3-configs/issues/1298) !

20260506 120953.284 ERROR            PET0055 No iterations performed in flux_atmocn_mod

20260506 120953.322 ERROR            PET0253 ERROR: 324 nans found in So_u
20260506 120953.323 ERROR            PET0253 ERROR: 324 nans found in So_v

dougiesquire:
We need to get to the bottom of this Nan issue...

anton-seaice:
Ill reconfirm that it also crashes without setting `--mca mpi_yield_when_idle true`

anton-seaice:
Yes - without this change it also crashed at ` atm : model date     19000103       40200`

dougiesquire (approved):

PR #1436, opened 2026-06-18
Set `PARALLEL_RESTARTFILES=True` as in other configs

dougiesquire:
All dev configs have `PARALLEL_RESTARTFILES=True` except for `dev-MC_100km_jra_iaf+wombatlite`. This doesn't actually do anything in the 100km configs but it's nice to have consistency.

I've checked that the parameter doc files are correct.


anton-seaice:
!test repro

anton-seaice (approved):

Issue #1463, opened 2026-06-25
Set up `M_regional_template` branch


dougiesquire:
In a [recent TWG meeting](https://forum.access-hive.org.au/t/cosima-twg-meeting-minutes-2026/5906/13#p-22820-regional-stencil-naming-convention-2) we discuss a desire to add/update a branch to `access-om3-configs` that will act as a “stencil” for regional configurations.

This will act as a template, but also will serve as something we can test against when we make updates/test upstream changes.

We agreed upon the name `regional_template` but I will add a `M_` prefix as this will only include the ocean. I will create this from the existing, though out-of-date, `helen/access-rM-tas5km_jra_iaf+wombatleg` branch and start a PR with the necessary changes.

PR #1493, opened 2026-07-03
Set up `M_regional_template` branch

dougiesquire:
**Summary**:

This PR is to set up a “template” branch for regional configurations (`M_regional_template` is currently the same as `helen/access-rM-tas5km_jra_iaf+wombatleg`). This branch will act as a stencil for the `regional-mom6` package, but also will serve as something we can test against when we make updates/test upstream changes. 

@ashjbarnes, sorry it took me so long to get to this. This branch runs with the latest ACCESS-OM3 exe. However, there are still some things that will need to be addressed before this can be merged:
- [ ] The input files used by `helen/access-rM-tas5km_jra_iaf+wombatleg` are no longer available. Some have been deleted and others have been moved into a “`to_delete `" folder. So for now, I’ve instead just used a copy of inputs @ashjbarnes had at `/g/data/tm70/ab8992/rom3-boundary-troubleshooting/rom3-run` (luckily I made a copy because these also no longer exist).
- [ ] I haven’t really edited the `MOM_input` settings other than to turn off generic tracers and a few other small changes. The `MOM_input` will need to be reviewed and maybe updated.
- [ ] I have used @ashjbarnes' `MOM_override` from `/g/data/tm70/ab8992/rom3-boundary-troubleshooting/rom3-run`. The `MOM_override` will need to be reviewed and maybe updated.
- [ ] The config uses a very simple `diag_table` with some grid metrics and depth, temp, salt, horizontal velocities on native grid. This is maybe fine given the purpose of the config.
- [ ] A `metadata.yaml` file should be added.
- [ ] Once ready, we'll need to update the checksums

@ashjbarnes and @helenmacdonald, you'll need to decide what to do about the input files. These need to be available in `vk83` with provenance information in their metadata so that they can be regenerated.

Please feel free to push changes to this PR (e.g. to update `MOM_input`, `MOM_override` etc)

**Issues Addressed:**

- #1463


ashjbarnes:
Awesome thanks so much @dougiesquire !!

ashjbarnes:
I've tested and run the model with the latest rmom6 (post a lot of changes that @manishvenu and I made at NCAR) and these template files, so all is well. A few thoughts in order of the bullet points above:

- I'll put the Tasmania input files somewhere in vk83 for testing purposes. Importantly though, the intention is for users to generate all input files when learning to use rmom6 with the demo, so these input files will only be used for ACCESS-NRI testing
- For MOM_input, shall we start from NCAR's MOM_input parameterisation choices for regional? They've done a lot more optimisation for generic regional configs recently so we could leverage what they've learned
- MOM_override is always generated by rmom6, so we shouldn't update it manually if this repo is used to test rmom6. If there are any changes due to rmom6 changes, I'll push these. 
- For the diag table: even this simplified one produces a lot of files. For a small regional run, could we have multiple variables in a single file like 3d_weekly or something? Or is that not the ACCESS-NRI way? I'd also change the outputs to weekly not monthly since for the demo we probably only want users running for like 10 days at a time

One last thing: using JRA surface forcing with GLORYS boundary forcing is a very unlikely choice for science runs due to the mismatch in data product. My preference would be that the rmom6 template used ERA5 surface forcing. I could do this now with a modified `fd.yaml` and `datm_in` file so we leverage rmom6's inbuild ERA5 functionality (it just mimics JRA forcing fields), or if the ERA5 forcing branch is ready and stable we could try that again.

ashjbarnes:
As discussed with @helenmacdonald and @dougiesquire , before merging we need to put rmom6 version used to generate each .nc file into the metadata of each .nc file. Ashley to do this. 

Quick changes:
 - Update diag_table to weekly output (maybe even daily for surface fields so user can at least see a bit of a timeseries for short runs? @dougiesquire ?)
 - Copy rmom6 demo notebook into this repo. This notebook is to be manually kept up to date with the rmom6 repo one by ACCESS-NRI by copying the latest version, at the same time updating the `config.yaml` to point to the newly generated .nc input files. This doesn't affect rmom6 since the config.yaml is overwritten to point to user's .nc files, but important for NRI testing. @dougiesquire did I explain that right?

Other items will be done as later PRs, including
 - Update the MOM_input to match [NCAR's](https://github.com/ESCOMP/MOM_interface/pull/316) that @manishvenu has worked on. @helenmacdonald to do this while looking into inputs for superaus
 - Include ERA5 forcing, hopefully using @ezhilsabareesh8 's method as long as there's no longer significant slowdown on small regional domains when trying to read global datasets. If not we can still use rmom6's era5 method

ashjbarnes:
Having thought about it more, I'm not a fan of having a copy of the notebook in this repo. The issue is that if the workflow is:

1. Get the notebook from regional-mom6 repo, where it's embedded in the docs
2. Follow the notebook which then tells you to...
3. git clone the template

then the user will now have just cloned another notebook almost identical to the first one into the run directory but that they have no use for. I think this is messy and confusing for new users.

@helenmacdonald , @dougiesquire can we think of another way to do this? Preferably this could be something done quietly in the background and not muddy things for users. Perhaps whoever maintains this template at ACCESS NRI can copy the notebook to the same folder as where the rmom6-generated forcing files will live for your OM3 testing?

ashjbarnes:
I've now implemented the appending of rmom6 version to the .nc metadata in this PR:

https://github.com/COSIMA/regional-mom6/compare/deploy_rom3?expand=1

As this was the only other thing to handle, my preference would be to get this merged into the main template and handle the above issue in a separate PR

ashjbarnes:
Another thing that needs fixing that's not really in my wheelhouse is payu checking the manifests:

After a fresh `git clone` of this branch, running through the regional-mom6 workflow to the end, which modifies the config.yaml to point to the newly generated input files, I tried the usual `payu setup` in the run directory and got this error, followed by a long list of all the new input files with mismatched hashes

```
laboratory path:  /scratch/gb02/ab8992/access-om3
binary path:  /scratch/gb02/ab8992/access-om3/bin
input path:  /scratch/gb02/ab8992/access-om3/input
work path:  /scratch/gb02/ab8992/access-om3/work
archive path:  /scratch/gb02/ab8992/access-om3/archive
Metadata and UUID generation is disabled. Experiment name used for archival: rOM3-tassie-demo
payu: work path already exists.
      Sweeping as --force option is True.
Removing work path /scratch/gb02/ab8992/access-om3/work/rOM3-tassie-demo
Removing symlink /g/data/gb02/ab8992/rom3-rundirs/rOM3-tassie-demo/work
payu: Found modules in /opt/Modules/v4.3.0
Loading access-om3/2026.05.002
  Loading requirement: access3-share/2026.03.002-m65pkhf access-cice/CICE6.6.3-1-uqsyhj7 access-mocsy/2025.07.002-5awgzcl
    access-generic-tracers/2026.05.000-7jsawsl access-mom6/2026.05.001-impfcxi access-ww3/2026.03.000-kjy7rmt
    access3/2026.03.002-7t6n2ft
Setting up access-om3
Checking exe, input and restart manifests
Run cannot reproduce: manifest manifests/input.yaml is not correct
Manifest path: stored hash != calculated hash
```

I've tried deleting the manifest folder entirely figuring payu might make a new one for me but that didn't work. I understand why there's all the checking of manifests, but this template necessarily needs to be easily modifiable by the user, since the point is that they'll re-generate all of the input files themselves

@aidanheerdegen @dougiesquire @anton-seaice @jo-basevi  any suggestions as to how to set things up so that the input files in config.yaml can be modified and user can then just run

```
payu setup
payu run -f
```

?

manodeep:
@ashjbarnes You have to set manifest->reproduce->input to False and then run payu setup. That should overwrite the stored hash

dougiesquire:
Yup, we should change that in the `config.yaml` for this config given how you're saying the config will be used

ashjbarnes:
Thanks! I did find the manifests section of the payu docs, but the "manifest options" section was pretty vague. Could these settings please be included there? I've made an [issue](https://github.com/payu-org/payu/issues/831)

ashjbarnes:
ok @helenmacdonald just for future reference we now have the "controversial era5 commit" to be undone when era5 forcing is done and someone has capacity to test it out on regional configs

ashjbarnes:
@helenmacdonald I'd rather not have the notebook in the repo. For users running the regular regional-mom6 workflow to generate new bespoke domains, they will be cloning this repo and end up with an unrelated notebook for the Tassie demo. Since this notebook isn't attached to `regional-mom6`, which will be updated independently of this template, you could have situations where a new user clones this repo and then finds a notebook that's _different_ than the new example notebook on `regional-mom6`

I'd like this repo to be as clean as possible and avoid confusion for new users. Since the only need for the notebook is provenance, could they just live on disk alongside where the inputfiles will be saved on vk83?

ashjbarnes:
What's the deal with the `testing` folder? This doesn't seem tied to the template at all - is this a placeholder for now which will be populated with checksums for the Tassie demo? 

If it's not to be implemented in the short term can we delete it and re-add later as a separate PR?

dougiesquire:
The `testing` directory contains checksums for testing reproducibility. The actual checksums in there are out of date and will need to be updated before we merge this PR. This is what my comment in the PR description is about, i.e.:

> * [ ]  Once ready, we'll need to update the checksums

ashjbarnes:
> The `testing` directory contains checksums for testing reproducibility. The actual checksums in there are out of date and will need to be updated before we merge this PR. This is what my comment in the PR description is about, i.e.:
> 
> > * [ ]   Once ready, we'll need to update the checksums

ahh ok thanks for explanation!

ashjbarnes:
@dougiesquire @helenmacdonald just checking what we're up to with this template? If the input files have been uploaded we should just need to reference the path in the `config.yaml` path rather than my directory

I'm going to make a hive post letting the community know that there's a usable version of rmom6 again (even if it's not gone through the full "release" procedure yet). I'll call it a pre-alpha, and accordingly will be pointing to the prerelease version of @atteggiani and @jo-basevi 's environment (it works great!!) 

But it would be nice if the workflow could point to the main template rather than this PR to keep things a bit less confusing

dougiesquire:
I think we should update to use the ERA5 datamode in CDEPS. I can help with that if needed.

Also, I thought we decided in the end that we would not include the notebook used to generate the inputs in this repo?

ashjbarnes:
Ok sounds good! Do you have time today @dougiesquire ? Maybe we could call and try test it out together?

Yes agreed, we need to delete the notebook and @helenmacdonald needs to point the config.yaml file to where she's uploaded the files

dougiesquire:
> Ok sounds good! Do you have time today @dougiesquire ? Maybe we could call and try test it out together?

I can make the changes if you can test? One issue is that I don't have access to your current inputs (in `gb02`). Could you please copy an `ls` of the contents of that directory so that I can check we have all the required inputs in `vk83`

ashjbarnes:
Sounds good! thanks Dougie. Below are the files that mom6 actually needs, there are some other residual files that come about when you generate these files, like weights, `bathymetry_unfinished.nc` and the (soon to be redundant) `xx_ERA5.nc` files

```
forcing_obc_segment_001.nc (and  2 and 3 and 4)
init_eta.nc
access-rom3-nomask-ESMFmesh.nc
access-rom3-ESMFmesh.nc
init_tracers.nc
vcoord.nc
bathymetry.nc
hgrid.nc
init_vel.nc
```

Perhaps we don't need the nomask mesh now either with the updated surface forcing? I can't remember

ashjbarnes:
You could also have a go at making the files yourself too if you're interested in trying regional-mom6? 

You can run the [demo](https://github.com/COSIMA/regional-mom6/blob/deploy_rom3/demos/ACCESS-rOM3-demo.ipynb) using [this environment](https://github.com/ACCESS-NRI/containerised-environments-infra/pull/34#issuecomment-5562566033), and it will generate all the input files again if all is set up right.

dougiesquire:
That notebook doesn't look complete?

Screenshot 2026-09-07 at 9 32 35 am

It would be good to remove this section from the notebook anyway if we're happy with the CDEPS ERA5 datamode. Let me make the change and you can test it out

ashjbarnes:
True, yeah I intend to get rid of tides in the basic demo. Just ignore those two cells and hopefully the rest of it runs fine. If not, good to know, I don't think anyone else but me (and maybe @helenmacdonald ?) have used the new containerised env workflow yet

dougiesquire:
I tried to update to the files that @helenmacdonald generated. But it seems that the mesh is inconsistent with the grid...

In the `config.yaml`, I changed:
```diff
input:
-    - /g/data/gb02/ab8992/rom3-inputdirs/access-rom3-demo
+    - /g/data/vk83/prerelease/configurations/inputs/access-rom3/regional_template/2026.08.14
```

and got:
```
FATAL from PE     2: ERROR: Difference between ESMF Mesh and MOM6 domain coords is greater than parameter EPS_OMESH. n, lonMesh(n), lon(n), diff_lon, EPS_OMESH=        1    144.4500000000000       -47.8190000000000             0.19227D+03          0.10000D-12
```

@helenmacdonald, did you try and run with these files?

ashjbarnes:
I can confirm that this branch in its current form works out of the box for me using my inputs (the ones that the `config.yaml`) points to. @helenmacdonald could you please confirm that the inputs you made and uploaded did work with this template?

dougiesquire:
@ashjbarnes, I've just pushed changes to update to using the ERA5 CDEPS datm datamode. I've tested that this runs for one month, but it would be good if you could also test.

Because I don't have access to `gb02` I wasn't able to update the inputs manifest at this point. We can do that when we transition to using input data on `vk83`.

dougiesquire:
> You can run the [demo](https://github.com/COSIMA/regional-mom6/blob/deploy_rom3/demos/ACCESS-rOM3-demo.ipynb) using [this environment](https://github.com/ACCESS-NRI/containerised-environments-infra/pull/34#issuecomment-5562566033), and it will generate all the input files again if all is set up right.

I'm trying a run with inputs generated fresh using this notebook and the `dev` environment linked here. If the run succeeds, can we please update the notebook to remove the now-redundant ERA5 forcing file generation? In addition to removing the cell calling `expt.setup_era5` it looks like the `setup_rOM3` call will also need updating?

Screenshot 2026-09-07 at 4 13 48 pm

Once the notebook is updated, I can use that to regenerate the inputs and then move those into `vk83`

helenmacdonald:
> @helenmacdonald, did you try and run with these files?

Sorry all – I am in a coastal-commons hackathon this week so don’t have much time to look at it. I am still flat-out on a few different meetings/activities in the coming weeks but I have some time in early October.

I did check that the files ran using the stencil produced by the regional-mom6 notebook  - but admittedly didn’t recheck after the metadata was added. Thanks for looking into this @dougiesquire. Just a reminder that not all files have the metadata automatically generated so some may need to be manually added.

ashjbarnes:
Yeah I think the metadata in the ESMFmesh files had become messed up. I noticed that the dimensions were swapped around according to ncdump, so I think more was modified than just adding provenance

dougiesquire:
I've pushed updates to the `MOM_input` to align with the `MC_panan4km_jra_ryf` config (see commit for details).

@ashjbarnes, I think we need to revisit the `diag_table`. I see you changed some diagnostics to 7 day means. I don't think this will play nicely with the default 1 month run length since 1 month is not generally divisible by 7 days.

ashjbarnes:
True! Yeah happy to change to 1,5 or 10 days.

ashjbarnes:
Ok as an update: @dougiesquire has fixed up the diag table and MOM_input

I've now merged rmom6 deployment branch into main, have tagged releases, and pointed the containerised env to the tagged versions of rmom6 and mom6_forge

Next week once the deployments trigger, Dougie can re-make all the input files with the main branch of rmom6, saving the version into the metadata. Then the files can be uploaded to vk83 and the config.yaml file pointed to them. We also need to then update the notebook to point to the pre-downloaded GLORYS data too, which then ought to be moved to vk83.

ashjbarnes:
Sorry team one last thing that I found: we'd partially renamed "vcoord" to "vgrid" a while back. vgrid makes more sense and I prefer it, it's more inline with the mom6_forge naming too. @helenmacdonald will just need to rename the `vcoord.nc` input to `vgrid.nc` in the template input files. Alternatively, just recreate them again with updated rmom6 so that the vertical grid file is called `vgrid.nc`

dougiesquire:
@ashjbarnes yeah we'll want the inputs in this config to be generated using the released version of regional-mom6 before we merge

ashjbarnes:
ok true! apologies I didn't find this earlier - we can merge the renaming PR quickly as it's just a find-and-replace job

The vgrid.nc naming convention also matches ACCESS-NRI better I think compared to global models

ashjbarnes:
With the updated MOM_input, I now get this warning:

```
WARNING from PE     0: set_diffusivity_init: SIMPLE_TKE_TO_KD can not be used reliably with USE_REGRIDDING.
```

Just thought I'd check if @dougiesquire knows what this is and if it's something we care about or not. Previously regridding was on but simple_tke_to_kt was not

panan is set up the same way though so it's probably fine

dougiesquire:
Yeah this is expected, at least in the sense that we get this in all our configs. The GFDL OM5 configuration also uses this parameter combo. Weirdly, the parameter description seems to contradict the warning, at least at face value:

```
SIMPLE_TKE_TO_KD = True         !   [Boolean] default = False
                                ! If true, uses a simple estimate of Kd/TKE that will work for arbitrary
                                ! vertical coordinates. If false, calculates Kd/TKE and bounds based on exact
                                ! energetics for an isopycnal layer-formulation.
USE_REGRIDDING = True           !   [Boolean] default = False
                                ! If True, use the ALE algorithm (regridding/remapping). If False, use the
                                ! layered isopycnal algorithm.
```

I think it's okay to leave for now and if we revisit in the other configs we can also update the template.

ashjbarnes:
Great! I definitely agree to keep it and revisit this stuff later, was just double checking

ashjbarnes:
Suggestion:

Remove the memory specification in config.yaml. My understanding is that this way the amount of memory will be requested that aligns with the number of nodes chosen. If left hardcoded users will need to remember to update this once PEs are scaled up

dougiesquire:
Fine with me. Feel free to push the change

ashjbarnes:
@aekiss here's the regional template branch. I understand that after chatting with @AVEllepola you had some thoughts about the remapping algorithms? Should we consider your suggestions in the generic regional template, or is it more a case-by-case basis and we should keep defaults?

ashjbarnes:
Inputs are on vk83 and I've just changed the config.yaml file to point to them. Ready for the checksum/manifest sorcery now

dougiesquire:
Thanks @ashjbarnes, could you please add a `metadata.yaml` file to the branch. You can see one of the other branches for what it should contain and how it should be formatted. E.g.

https://github.com/ACCESS-NRI/access-om3-configs/blob/dev-MC_25km_jra_ryf/metadata.yaml

Issue #404, opened 2024-09-03
Calving flux via iceberg distribution map


aekiss:
Do we want to distribute the calving meltwater and heat flux according to iceberg distribution? This was done in ACCESS-CM2, but not in the ACCESS-OM2 release (although Pedro Colombo (https://github.com/pedrocol/basal_mom5-collaborative-project) and Jingwei Zhang have explored adding it).

Pedro has also investigated distributing basal and iceberg melt at depth.

aekiss:
related: https://github.com/ACCESS-NRI/access-om3-configs/issues/1010#issuecomment-3672947327
and https://github.com/ACCESS-Community-Hub/access-om3-paper-1/issues/20#issuecomment-3600325477

aekiss:
This has become a top priority to implement, to avoid excessive sea ice buildup and CICE crashes (see links in previous post).

@kieranricardo how is iceberg spreading implemented in ACCESS-CM3? Perhaps we can directly adopt that for OM3?

anton-seaice:
I think it's spread in the same way as the liquid runoff and done on the atmosphere grid.

See 

https://github.com/ACCESS-NRI/dev_coupling/blob/main/scripts/runoff_smoothing_weights.py

anton-seaice:
I think i have space to look at this now - ill begin working on it. 

Ill investigate spreading the volumne of "frozen runoff" from JRA55do licalvf field according the pattern a 12-month climatological pattern from Merino 2016 and see how that looks

anton-seaice:
> There's a fair bit of temporal variability in [Merino et al 2016 supp 2](http://dx.doi.org/10.1016/j.ocemod.2016.05.001).

This only covers Antarctica - is there a preferred pattern for Greenland ?

chrisb13:
@claireyung organised a meeting this morning to discuss Australian ice shelf developments (@chrisb13, Alexander Huth, Nicole Schlegel, Olga Sergienko, Bob Hallberg, @AndyHoggANU, @angus-g present). Iceberg representations came up. 

I realise this is late in this discussion but just to mention that we were told that it is possible to use the mom6 iceberg code in the absence of SIS2 (i.e. CICE and nuopc). The papers to look at are (I gather there's quite a few other papers):
 - https://www.sciencedirect.com/science/article/pii/S146350031000065X
 - https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2017MS001002

[...]

PR #1503, opened 2026-07-08
docs for iceberg spreading

anton-seaice:
Some docs for #404


documentation/docs/pages/inputs/Forcing-data-models.md
```diff
@@ -39,7 +39,10 @@ Input data is first interpolated to the model time within DATM, second remapped
...
+
+- For *liquid runoff*, the differences in landmask between the incoming data on the JRA55-do grid and the OM3 grids can cause runoff to be remapped to land cells or on cells away from the coastlines. When runoff would be placed on land cells or non-coastal ocean cells, the volume of runoff is crudely moved to the nearest coastal ocean cell using pre-generated weights from the [generate_rof_weights.py](https://github.com/ACCESS-NRI/om3-scripts/blob/main/mesh_generation/generate_rof_weights.py) script. _generate_rof_weights.py_ selects the nearest coastal ocean cell by a BallTree algorithm using Haversine distances (i.e. the shortest disance on a sphere). The combined effect of the two remapping steps is that the full global volume of runoff enters the ocean, and only at cells which about land cells.
```
dougiesquire: "about" -> "abut"

```suggestion
- For *liquid runoff*, the differences in landmask between the incoming data on the JRA55-do grid and the OM3 grids can cause runoff to be remapped to land cells or on cells away from the coastlines. When runoff would be placed on land cells or non-coastal ocean cells, the volume of runoff is crudely moved to the nearest coastal ocean cell using pre-generated weights from the [generate_rof_weights.py](https://github.com/ACCESS-NRI/om3-scripts/blob/main/mesh_generation/generate_rof_weights.py) script. _generate_rof_weights.py_ selects the nearest coastal ocean cell by a BallTree algorithm using Haversine distances (i.e. the shortest disance on a sphere). The combined effect of the two remapping steps is that the full global volume of runoff enters the ocean, and only at cells which abut land cells.
```

dougiesquire (approved):
Thanks @anton-seaice

PR #1536, opened 2026-08-04
Update doc folder in main branch with key config files to make searching easier

### LINKED ISSUE #402: Regularly run MOM6-CICE6/doc/update_doc.sh
From https://github.com/COSIMA/access-om3/issues/75#issuecomment-1943051644 - this is needed to keep the files linked to from the [wiki](https://github.com/COSIMA/access-om3/wiki) up to date.

> We should run [MOM6-CICE6/doc/update_doc.sh](https://github.com/COSIMA/MOM6-CICE6/blob/main/doc/update_doc.sh) regularly to keep the files in [MOM6-CICE6/doc](https://github.com/COSIMA/MOM6-CICE6/blob/main/doc) up to date with the `1deg_jra55do_ryf` branch (or perhaps a different branch in future).

> At some point we may want to switch to basing the wiki documentation on MOM6-CICE-WW3, so this PR will facilitate that https://github.com/ACCESS-NRI/access-om3-wav-configs/pull/16

anton-seaice:
Closes #402

update the doc folder in main branch with key config files, so that searchable links can be used

example:

https://github.com/search?q=repo%3AACCESS-NRI%2Faccess-om3-configs+DT_THERM+path%3A**%2FMOM_input&type=code


anton-seaice:
hey @CodeGat - i've been badly writing CI, would you like to review it ? 🥲

.github/workflows/update-doc-folder.yml
```diff
@@ -0,0 +1,48 @@
...
+  update-doc-folder:
+    runs-on: ubuntu-latest 
+    permissions:
+      contents: write
+
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          sparse-checkout: doc
+      - name: Import Commit-Signing Key
```
CodeGat, line 20: If you do: 

```suggestion
      - uses: actions/checkout@v7
        with:
          sparse-checkout: doc
          depth: 0

      - name: Import Commit-Signing Key
```

you can delete https://github.com/ACCESS-NRI/access-om3-configs/pull/1536/changes#diff-3b9c2e535fea60e0f146c871a5f0ca8a67598b809a9e81ada73400d30a4b7bb6R33 I think!

.github/workflows/update-doc-folder.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: update doc folder from config branch
+        run: |
+          git fetch origin ${{ env.config-branch-name }}
+          commit=$(git rev-parse --short origin/${{ env.config-branch-name }})
+          git checkout origin/${{ env.config-branch-name }} ${{ env.doc-files }}
+
+          for file in ${{ env.doc-files }}
+          do
+              mv ${file} doc
+              git restore --staged ${file}
```
CodeGat: I'm a bit confused by this part specifically - so we're getting files from the `config-branch-name`, moving them into `docs/` (which is fine), but then unstaging the original file paths - wouldn't they not exist, because they're moved?

CodeGat (commented):
Just a couple of comments!

.github/workflows/update-doc-folder.yml
anton-seaice: Oh yeah, probably fine to skip this line. I thought there was a name overlap, but there isn't

.github/workflows/update-doc-folder.yml
anton-seaice: how does sparse checkout interact with depth ?

like does it only fetch doc for every branch ?

anton-seaice: nvm, i remembered

.github/workflows/update-doc-folder.yml
```diff
@@ -0,0 +1,49 @@
...
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+          git_tag_gpgsign: true
+      - name: update doc folder from config branch
+        run: |
+          # figure out latest commit in config-branch
+          commit=$(git rev-parse --short origin/${{ env.config-branch-name }})
+
+          # checkout config files from the config-branch
+          git checkout origin/${{ env.config-branch-name }} ${{ env.doc-files }}
```
CodeGat, line 37:
```suggestion
          git restore --source=origin/${{ env.config-branch-name }} -- ${{ env.doc-files }}
```

I looked up what `git checkout origin/BRANCH -- FILE` actually does and it does stage the changes initially...but `git restore --source=origin/BRANCH -- FILE` doesn't! So initially you were right to have the `git restore $file` - sorry mate! At least this makes it a one-liner...

anton-seaice: Oh yeah, of course

anton-seaice:
Should be ready to go @CodeGat

CodeGat (approved):

Issue #949, opened 2025-12-03
Restart reproducibility failure in MCW configs


ezhilsabareesh8:
### Issue description
Restart-repro tests are failing for MCW configs (checksums differ across many fields, refer [here](https://github.com/ACCESS-NRI/access-om3-configs/pull/841#issuecomment-3430554438)). This is likely tied to the wave–ice fracture algorithm, which uses a random component `wave_spec_type = random` in `ice_in`, relevant discussion is [here](https://github.com/ACCESS-NRI/access-om3-configs/issues/283#issuecomment-2750054143) and [here](https://github.com/ACCESS-NRI/access-om3-configs/pull/841#issuecomment-3604407039).

There were earlier discussions about storing the random seed in the restart so the run becomes reproducible, but it’s not implemented yet.

### Issue checklist

 
- [x] I have labelled this issue with the affected configurations (e.g. `MC_25km_jra_ryf`, [1] more options)
- [x] I have added the issue information to the [ACCESS-OM3 Mega Project board!](https://github.com/orgs/ACCESS-NRI/projects/24):
After issue creation, under _Projects_ please set:
 - [x] Status (blocked, Todo, In Progress, Done)
 - [x] Priority (P0, P1, P2) 
 - [x] Component

ezhilsabareesh8:
Restart reproducibility works as expected when `wave_spec_type = none`. See this comment for reference: https://github.com/ACCESS-NRI/access-om3-configs/pull/841#issuecomment-3604510266

PR #1542, opened 2026-08-05
Set CICE wave spectrum type to constant

ezhilsabareesh8:
**1. Summary**:

This change sets `wave_spec_type = "constant"` in CICE as a temporary workaround for the restart reproducibility failure in MCW configurations ([#949](https://github.com/ACCESS-NRI/access-om3-configs/issues/949)). It should be superseded once Noah’s ICDR method is implemented in Icepack.

**2. Issues Addressed:**

- https://github.com/ACCESS-NRI/access-om3-configs/issues/949

**5. CI Testing**

- [x] `!test repro` or `!test repro commit ` has been run

**6. Reproducibility**

- [ ] Yes
- [x] No - Science changes

**7. Documentation**

- [ ] Yes
- [x] N/A

- [ ] Yes: 
- [x] N/A

**8. Formatting**

- [ ] Yes
- [x] N/A

**9. Merge Strategy**

- [x] Merge commit
- [ ] Rebase and merge
- [ ] Squash

### DIFF
--- a/ice_in
+++ b/ice_in
@@ -67,7 +67,7 @@
   tfrz_option = "linear_salt"
   update_ocn_f = .true.
   ustar_min = 0.0005
-  wave_spec_type = "random"
+  wave_spec_type = "constant"
   /
 &domain_nml
   block_size_x = 15


anton-seaice (commented):
Just checking the motivations here:

The change is proposed just for restart reproducability, but makes results worse ?

ezhilsabareesh8:
Yes, that’s right. The `constant` method is less accurate, but the `random` method has some computational challenges. @NoahDay may have a useful reference that discusses these issues.

anton-seaice:
In my view you can leave this as-is, and just note it as a limitation in the release note (and recommend people don't change the run length)

NoahDay:
For reference, in our MCW testing `random` increased the overall computational cost by 3-fold (~12 kSU per year; ~15 hours/model year), whereas `constant` was ~4 kSU per year (~5 hours/model year). (Both tests were completed without any of the recent load balancing or performance improvements.)

Similar findings have been reported in [Horvat and Roach (2022)](https://gmd.copernicus.org/articles/15/803/2022/#Ch1.S4) that report that the implementation of `random` into CICE increases computation times by an order of magnitude. At CESM meetings I've also seen similar computational costs.

We're working on a new `wave_spec_type` option but since this is quite a big scientific change, I think it's probably best not to include it in the release(?)

anton-seaice:
Are the results with `constant` ok / useable ? e.g. that's the data will end up sharing

NoahDay:
`constant` still results in wave-induced breakup which varies spatio-temporally (it's just that the phase of the spectral wave components are all the same). 

Here I've plotted the mean floe sizes from other studies and MCW for `constant` (second to last column) and `random` (last column). The MCW runs aren't like-for-like (different time periods, mixing, grids, forcing etc), but I think `constant` is a better transition from ice edge->MIZ->inner pack (bottom row), whereas `random` is likely producing excessive breakup, which I've also heard from the authors of these studies mention. (The new breakup method will also address the excessive breakup concern.)
unknown-21

anton-seaice:
!test repro commit

manodeep:
> For reference, in our MCW testing `random` increased the overall computational cost by 3-fold (~12 kSU per year; ~15 hours/model year), whereas `constant` was ~4 kSU per year (~5 hours/model year). (Both tests were completed without any of the recent load balancing or performance improvements.)
> 
> Similar findings have been reported in [Horvat and Roach (2022)](https://gmd.copernicus.org/articles/15/803/2022/#Ch1.S4) that report that the implementation of `random` into CICE increases computation times by an order of magnitude. At CESM meetings I've also seen similar computational costs.

Can someone point me to the source code pathways that execute for both `random` and `constant`?

anton-seaice:
!test repro commit

anton-seaice:
> Can someone point me to the source code pathways that execute for both `random` and `constant`?

https://github.com/CICE-Consortium/Icepack/blob/0fa79f8a8c35798adc8e177c25ff7ea5857bb70f/columnphysics/icepack_wavefracspec.F90#L452

I think the wave_frac subroutine would be called for every grid point

anton-seaice:
!test repro commit

NoahDay:
Yep, it's called whenever there's "sufficient" ice coverage and wave height. The `random` option is stochastic and requires multiple iterations to converge (making it costly and not reproducible). `constant` is deterministic (reproducible and doesn't require the iterations), but lacks some realism since ocean waves are inherently stochastic.

manodeep:
These [lines of code](https://github.com/CICE-Consortium/Icepack/blob/0fa79f8a8c35798adc8e177c25ff7ea5857bb70f/columnphysics/icepack_wavefracspec.F90#L476-L542) could certainly use a closer look with a profiler. Lots of reductions and divisions - might be possible to refactor and improve

anton-seaice:
Looks good to me, slightly surprised this isn't answer changing in the repro tests ?

ezhilsabareesh8:
`wave_frac` is only called where there is ice (aice > 0.01), significant wave height exceeds 0.1 m, and the FSD is not already entirely in the first floe-size category, the the activation conditions is [here](https://github.com/CICE-Consortium/Icepack/blob/daa41638c6cef298583f35ee4dec5ab1bb077aea/columnphysics/icepack_wavefracspec.F90#L269). The CI run is only 6 hours and WW3 spins up from rest, so it may not encounter those conditions.

anton-seaice (approved):

NoahDay:
Also, the FSD will be initialised according to a power-law ([code](https://github.com/CICE-Consortium/Icepack/blob/0fa79f8a8c35798adc8e177c25ff7ea5857bb70f/columnphysics/icepack_fsd.F90#L329-L344)) meaning that sea ice will mostly be limited to smaller floe sizes (less prone to breakup)

Issue #1003, opened 2025-12-15
1-year continuous MCW runs crash with memory error


ezhilsabareesh8:
### Issue description

When running the 100 km MCW configuration with the recent 100 km MC updates (PR: https://github.com/ACCESS-NRI/access-om3-configs/pull/841) as a continuous one-year run, the job tends to crash after around six months, either due to memory exhaustion or with exit code 137. This still happens even after significantly increasing the requested memory and using `normalsr` nodes. In contrast, running the same configuration in monthly chunks (e.g. `payu run -n 12`) is stable and has been tested for at least two years. The issue occurs with both the current and older run sequences, while earlier versions were stable, which suggests that recent changes may be causing memory to build up over time, possibly due to a memory leak or accumulation issue during long continuous runs.

ezhilsabareesh8:
Update: the WW3 memory growth in [`MCW_100km_era5_iaf_KPP`](https://github.com/ACCESS-NRI/access-om3-configs/tree/MCW_100km_era5_iaf_KPP) has now been traced to the ST6 source-term routines and a fix has been verified.

The fix is in this WW3 commit:

https://github.com/ACCESS-NRI/WW3/commit/9acadda871a263cdb00ddf979596e72055640099

The test used the ACCESS-OM3 build discussed here:

https://github.com/ACCESS-NRI/ACCESS-OM3/pull/109#issuecomment-4795378479

#### What was causing the memory growth

The issue was caused by repeated calls to `IRANGE()` inside frequently-called ST6 source-term routines:

* `W3SIN6`
* `W3SDS6`
* `W3SWL6`
* `LFACTOR`

`IRANGE()` returns an allocatable array. These calls were happening inside the sea-point/source-term loop, so even though each individual allocation is small and temporary, the repeated allocate/deallocate cycle was enough to steadily increase the resident memory of the WW3 ranks over long continuous runs.

This is why the problem only became obvious in long runs: the model was not retaining one large obvious array, but was repeatedly exercising the heap in a very tight loop. Over time this caused monotonic RSS growth on the WW3 ranks and eventually led to the OOM failure.

The source-term diagnostic showed that the growth was concentrated in the ST6 wind-input and dissipation terms:

```text
DISSIPATION  ~71%
WIND_INPUT   ~29%
other source-term groups ~0%
```

#### Changes

The fix removes the repeated allocatable-function calls from the hot loop.

[...]

PR #1578, opened 2026-08-14
Set run length to nyears

ezhilsabareesh8:
**1. Summary**:

Changed the run length to `nyears` and wall-time accordingly


anton-seaice (approved):
#1003 fixed a memory leak, so this works now

